PR #10, opened 2022-03-14
Use batch logic to fetch NFTs

RustySol:


src/api/fetchNftsFromWallet.ts
```diff
@@ -1,50 +1,73 @@
...
   const tokenAccounts = await connection.getParsedTokenAccountsByOwner(
     walletAddress,
     { programId: TOKEN_PROGRAM_ID }
   );
 
-  console.log('fetchNftsFromWallet: token accounts', tokenAccounts);
-  for (let i = 0; i < tokenAccounts.value.length; i++) {
-    const tokenAccount = tokenAccounts.value[i];
-    const tokenAmount = tokenAccount.account.data.parsed.info.tokenAmount;
+  return fetchDataArrayInBatches(
```
haVincy, line 16: `fetchDataArrayInBatches` should be a general method for other fetch caller to use the data and cook it to return to the UI component, should not coupled with the fetch caller.

src/api/fetchNftsFromWallet.ts
```diff
@@ -1,50 +1,73 @@
...
-      tokenAmount.amount === '1' || tokenAmount.amount === 1;
-    const tokenDecimalsIsZero =
-      tokenAmount.decimals === '0' || tokenAmount.decimals === 0;
+const processBatch = (
+  aggregated: SingleTokenInfo[],
+  batch: SingleTokenInfo[]
+) => {
+  aggregated = aggregated.concat(batch);
+  return aggregated;
+};
```
haVincy, line 40: It seems we don't really handle the single batch data here? It still `concat` each batch and return the entire result to the caller. If we are not ready to handle this, I suggest we just remove it, keep the `fetchDataArrayInBatches` in original. 

If we want to handle this as a library usage in this API call, we should let single batch callback as optional function parameter in `fetchNftsFromWallet` to let user to decide whether want to handle the NFT datas by batch to render UI progressively or want to wait them all finished and render the UI.

RustySol: agree, I will remove the single batch callback for now

PR #29, opened 2022-03-31
Fix wrong programId, enhance tokenAccount metadata handling

RustySol:


haVincy (approved):

src/api/fetchMetadata.ts
```diff
@@ -68,27 +68,32 @@ export const singleTokenInfoPromise = async (
...
+      })
+      .catch((err) => {
+        console.log('failed to fetch uri data');
+        console.log(err);
+        return null;
+      });
+  } else {
+    console.log(`tokenAccount ${tokenAccountAddress} does not have metadata`);
+    return null;
+  }
```
haVincy, line 98: I just realized this is in `singleTokenInfoPromise`, can we change to do following

```ts
if (tokenInfo) {
    const dataUriResult= await safeAwait(axios.get(tokenInfo.data.uri));
     if (dataUriResult.error) {
         throw Error(dataUriResult.error)
     }
     const nftUriData = dataUriResult.data as any;
     nftImage = nftUriData.image;
     nftAnimation = nftUriData.animation_url;
     nftDescription = nftUriData.description;
     return {
         tokenAccountAddress: tokenAccountAddress,
         metadata: tokenInfo,
         edition: tokenEdition,
         tokenMintAddress: token.mint.toString(),
         nftImage,
         nftAnimation,
         nftDescription,
     };
} else {
   console.log(`tokenAccount ${tokenAccountAddress} does not have metadata`);
    return null;
}
```

PR #40, opened 2022-04-07
fix: config story book with less loader

vanminh1701:


haVincy (approved):

.storybook/preview.js
```diff
@@ -1,5 +1,5 @@
-import '../dist/index.css';
-
+// import '../dist/index.css';
```
haVincy, line 1: ditto.

PR #41, opened 2022-04-07
[LIQ-424] Sort orders

ahsan8244:
Screenshot 2022-04-08 at 2 44 52 AM

The photo shows an outdated screenshot, I made the selected label's text purple in the latest commit

### DIFF
--- a/src/api/backend/OrderAPI.ts
+++ b/src/api/backend/OrderAPI.ts
@@ -1,11 +1,25 @@
 import { ListBase, Order, SingleBase } from 'solana-candy-shop-schema/dist';
 import axiosInstance from '../../config/axiosInstance';
+import qs from 'qs';
+
+export type SortBy = {
+  column: string;
+  order: 'asc' | 'desc';
+};
 
 export async function fetchOrdersByStoreId(
-  storeId: string
+  storeId: string,
+  sortBy?: SortBy
 ): Promise<ListBase<Order>> {
+  const queryObject = {} as any;
+  if (sortBy) {
+    queryObject['orderByArr'] = JSON.stringify(sortBy);
+  }
+
+  const queryString = qs.stringify(queryObject);
+
   return axiosInstance
-    .get<ListBase<Order>>(`/order/${storeId}`)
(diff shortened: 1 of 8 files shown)


src/core/Orders.tsx
```diff
@@ -24,13 +56,17 @@ export const Orders: React.FC<OrdersProps> = ({
...
 }) => {
+  const [sortedByOption, setSortedByOption] = useState(SORT_OPTIONS[0]);
   const [orders, setOrders] = useState<any[]>([]);
   const [loading, setLoading] = useState(false);
 
   // handle fetch data
   useEffect(() => {
     setLoading(true);
-    fetchOrdersByStoreId(candyShop.candyShopAddress().toString())
+    fetchOrdersByStoreId(
```
ahsan8244, line 66: Why have we not been using `candyShop.orders()` here?

ahsan8244:
Just a terminology nit, I have used `sort` instead of `order`. Because of our association of the word "orders" with NFT orders, it is probably best to use the word "sort" for sorting purposes.

src/core/Orders.tsx
RustySol: Agree this we should use candyshop.orders()

src/api/backend/OrderAPI.ts
```diff
@@ -1,11 +1,25 @@
...
+};
 
 export async function fetchOrdersByStoreId(
-  storeId: string
+  storeId: string,
+  sortBy?: SortBy
 ): Promise<ListBase<Order>> {
+  const queryObject = {} as any;
+  if (sortBy) {
+    queryObject['orderByArr'] = JSON.stringify(sortBy);
```
RustySol, line 16: does `sortBy` need to be an array?

ahsan8244: A bit confused about this. When I tried sending an array, the backend is unable to parse it. Seems like the only thing that works is an object

RustySol: noted

RustySol (approved):

PR #45, opened 2022-04-08
[Perf/Chore] Speed up hot reload time

ahsan8244:
# Initial Improvements

This speeds up the hot reload time from about 30s to 10s. The main steps taken to achieve this are:
1. upgrade parcel to version 2.
2. reorganise repo into yarn workspaces, and move parcel's root to repo root. This is needed because parcel 2 cannot detect changes in linked packages unless they are within the parcel root. Also, parcel 2 cannot detect changes outside its root, which is what we were relying on with parcel 1. 

# EDIT: Port project to Vite 

I ported the whole project to Vite instead. Getting instant hot reload, and blazing fast builds now. 

TODO: 
~1. Update readme with dev setup instructions~
2. Ensure all production build and deploy commands work with this structure
~3. Fix pre-push hooks~

# EDIT 2: Rollback to Parcel 2 + Rollup

Facing hella issues with Vite's production build for the library. Dev mode and example build work perfectly, but we cannot switch to Vite unless we can build the library properly. 

Update: excluding external deps from rollup build.

TODO: 
1. Update readme with dev setup instructions
2. Ensure all production build and deploy commands work with this structure
3. Fix pre-push hooks


ahsan8244:
@beepsaidblue need your help to test this out

.husky/pre-push
```diff
@@ -1,4 +0,0 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-
-yarn format:check
```
haVincy, line 4: Why we remove this, doesn't it will remove the pre-push git hook that injected by `husky`?

ahsan8244: fixing this is in my todos in this PR's description. Will do that before we merge 💯

ahsan8244: @haVincy also btw, can u advise on how this will affect storybook? How can we ensure storybook works even after this is merged?

haVincy: I'm not familiar with Storybook , just seeing you remove the main file from it?

PR #52, opened 2022-04-10
[fix] Fix stroybook's confusing error msgs

haVincy:
When running `yarn build` or `yarn start`, current storybook relevant files will have errors that is confusing from other real error from candy-shop. This PR will eliminate those errors by properly extending the mock class. 

Let me know if anything missed or misplacement.


src/core/Sell.tsx
```diff
@@ -41,7 +41,7 @@ export const Sell: React.FC<SellProps> = ({
           const [userNfts, sellOrders] = await Promise.all([
             fetchNftsFromWallet(connection, walletPublicKey),
             fetchOrdersByStoreIdAndWalletAddress(
-              candyShop.candyShopAddress().toString(),
+              candyShop.candyShopAddress.toString(),
```
haVincy, line 44: This is the getter usage for reference.

PR #54, opened 2022-04-11
Decouple AnchorWallet from CandyShop instantiation

RustySol:


example/CandyShopContent.tsx
```diff
@@ -49,7 +48,7 @@ export const CandyShopContent: React.FC = () => {
         <h1 style={{ textAlign: 'center' }}>Sell Your NFTs</h1>
         <Sell
           connection={connection}
-          walletPublicKey={wallet?.publicKey}
+          wallet={wallet}
```
haVincy, line 51: I think we should keep `wallet.publicKey` as param if `Orders` and `Sell`only need the publicKey not entire `wallet` obj, let `CandyShopContent` to handle if `wallet` is `undefined`,  it should not pass it to the children but show the connect wallet.

Can reuse LIQNFT's connect button component.

- https://github.com/LIQNFT/liquid-frontend/blob/master/src/components/ConnectButton/index.tsx

src/components/BuyModal/index.tsx
```diff
@@ -14,15 +15,15 @@ import './style.less';
 export interface BuyModalProps {
   order: OrderSchema;
   onClose: any;
-  walletPublicKey: PublicKey | undefined;
+  wallet?: AnchorWallet;
```
haVincy, line 18: ditto. keep the `walletPublicKey: PublicKey` , removed the undefined from optional type, restrict caller must make sure wallet is valid before calling `BuyModal`.

src/components/SellModal/index.tsx
```diff
@@ -17,12 +18,14 @@ export interface SellModalProps {
   onCancel: any;
   nft: SingleTokenInfo;
   candyShop: CandyShop;
+  wallet?: AnchorWallet;
```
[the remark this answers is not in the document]
RustySol: good point

example/CandyShopContent.tsx
haVincy: Let caller to handle it, we just make sure we can the `walletPublicKey` is `PublicKey` type not `undefined`.

haVincy:
Assigned to myself, will solve the error and push new patch.

lib/src/core/OrderDetail/index.tsx
```diff
@@ -155,30 +157,33 @@ export const OrderDetail: React.FC<OrderDetailProps> = ({
...
+              }
             >
               Buy Now
             </button>
           )}
         </div>
-        {step === 1 && (
-          <Modal onCancel={() => setStep(0)} width={600}>
+        {state === TransactionState.PROCESSING && (
+          <Modal onCancel={() => setState(0)} width={600}>
```
haVincy, line 176: Another missing update........will update in next push.

lib/src/utils/rc-notification.tsx
```diff
@@ -20,17 +20,21 @@ Notification.newInstance(
...
-          {type === 'success' && <IconSuccess />}
-          {type === 'error' && <IconError />}
+        <div className="cds-left">
+          {type === NotificationType.Success && <IconSuccess />}
+          {type === NotificationType.Error && <IconError />}
         </div>
-        <div className="candy-right">
-          {type === 'success' && <div className="title">Success</div>}
-          {type === 'error' && <div className="title">Error</div>}
+        <div className="cds-right">
```
haVincy, line 31: Mistake when rebasing, will put back to `candy-left` & `candy-right` in next push

lib/src/core/Sell/index.tsx
```diff
@@ -33,7 +34,7 @@ enum LoadStatus {
  * React component that allows user to put an NFT for sale
  */
 export const Sell: React.FC<SellProps> = ({
-  walletPublicKey,
+  wallet,
```
haVincy, line 37: Should also add the check for wallet as the same in Buy. Will update.

lib/src/utils/ErrorHandler.ts
```diff
@@ -0,0 +1,20 @@
+import { notification, NotificationType } from './rc-notification';
```
RustySol, line 1: nice one!

PR #58, opened 2022-04-12
[LIQ-429] Infinite scroll pagination

ahsan8244:


lib/src/api/backend/OrderAPI.ts
```diff
@@ -9,12 +9,20 @@ export type SortBy = {
 
 export async function fetchOrdersByStoreId(
   storeId: string,
-  sortBy?: SortBy
+  sortBy?: SortBy,
+  offset?: number,
+  limit?: number
```
haVincy, line 14: What if caller want to specify `offset` or `limit` ? It will force to set up the `sortBy` param.

I think we should add the obj for these options like
```ts
interface OrdersQuery {
   sortBy?: SortBy,
   offset?: number,
   limit?: number
} 

  storedID: string, 
  ordersQuery?: OrdersQuery): Promise<ListBase<Order>> {
   ....
}
```

lib/src/core/CandyShop.tsx
```diff
@@ -108,8 +108,17 @@ export class CandyShop {
     return this._programId;
   }
 
-  async orders(sortBy?: SortBy): Promise<ListBase<Order>> {
-    return fetchOrdersByStoreId(this._candyShopAddress.toString(), sortBy);
+  async orders(
+    sortBy?: SortBy,
+    offset?: number,
+    limit?: number
```
haVincy, line 114: ditto.

lib/src/api/backend/OrderAPI.ts
ahsan8244: good point, thanks!

lib/src/components/InfiniteOrderList/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+  loadNextPage,
+}) => {
+  return (
+    <InfiniteScroll
+      dataLength={orders.length}
+      next={loadNextPage}
+      hasMore={hasNextPage}
+      loader={
+        <Flex>
+          {Array(4)
```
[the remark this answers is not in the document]
ahsan8244: yes let me turn this into a constant

haVincy (approved):

PR #67, opened 2022-04-13
[fix] Fix concurrent calls in Sell.tsx

haVincy:


RustySol (approved):
Nice PR! Just curious, so what exactly cuz the concurrent calls here?

haVincy:
> Nice PR! Just curious, so what exactly caused the concurrent calls here?

Simply saying is `useEffect` to render twice with those `setState` to call Promises. Also the Promise can be in better place without nested too much.

PR #71, opened 2022-04-14
prevent duplicate instance candyShop

vanminh1701:


example/CandyShopContent.tsx
```diff
@@ -11,16 +11,34 @@ import {
...
+          new CandyShop(
+            new web3.PublicKey(CREATOR_ADDRESS),
+            new web3.PublicKey(TREASURY_MINT),
+            new web3.PublicKey(CANDY_SHOP_PROGRAM_ID),
+            'devnet',
+            wallet
+          )
+        ),
+      100
+    );
```
haVincy, line 36: Could we try not to use `setTimeout` to initiate the CandyShop? If the problem is wallet will be updated from `useAnchorWallet` 2 times to cause re-render and recreate the CandyShop, the solution here I feel should use `useRef` for `new CandyShop` obj ?

beepsaidblue: as part of this, can we look at how to minimize the API calls on the example site?

from what I observe, the Orders component is calling the same API several times (we want it to call once per load)

PR #73, opened 2022-04-14
Add shop currency / decimal, price and volume decimal customization

beepsaidblue:


lib/src/core/CandyShop/index.tsx
```diff
@@ -42,13 +42,22 @@ export class CandyShop {
...
+  private _volumeDecimals: number;
   private _program: Program | undefined;
 
   constructor(
     candyShopCreatorAddress: web3.PublicKey,
     treasuryMint: web3.PublicKey,
     candyShopProgramId: web3.PublicKey,
-    env: web3.Cluster
+    env: web3.Cluster,
+    currencySymbol: string,
```
RustySol, line 57: let's add `?` annotation here

lib/src/core/CandyShop/index.tsx
```diff
@@ -42,13 +42,22 @@ export class CandyShop {
   private _treasuryMint: web3.PublicKey;
   private _programId: web3.PublicKey;
   private _env: web3.Cluster;
+  private _currencySymbol: string;
+  private _currencyDecimals: number;
+  private _lamportsPerCurrency: number;
```
RustySol, line 47: I think can call it `baseUnitsPerCurrency`. Lamport is specific to SOL

example/CandyShopContent.tsx
```diff
@@ -19,7 +19,7 @@ export const CandyShopContent: React.FC = () => {
     new web3.PublicKey(CREATOR_ADDRESS),
     new web3.PublicKey(TREASURY_MINT),
     new web3.PublicKey(CANDY_SHOP_PROGRAM_ID),
-    env
+    env,
```
haVincy, line 22: Why need to add a comma here, looks like the linter option has to be corrected or something?

lib/src/components/BuyModal/BuyModalDetail.tsx
```diff
@@ -23,7 +23,12 @@ const BuyModalDetail: React.FC<BuyModalDetailProps> = ({
   candyShop,
 }) => {
   const orderPrice = useMemo(() => {
-    return (Number(order?.price) / web3.LAMPORTS_PER_SOL).toFixed(3);
+    return (
+      Number(order?.price) / candyShop.lamportsPerCurrency
```
haVincy, line 27: What if `order ` is invalid? Seems we will return the invalid value here.

lib/src/components/CancelModal/CancelModalDetail.tsx
```diff
@@ -43,6 +43,15 @@ export const CancelModalDetail = ({
 
   const buttonContent = 'Cancel listing';
 
+  const orderPrice = useMemo(() => {
+    return (
+      Number(order?.price) / candyShop.lamportsPerCurrency
```
haVincy, line 48: ditto.

lib/src/core/CandyShop/index.tsx
```diff
@@ -42,13 +42,22 @@ export class CandyShop {
...
+  private _priceDecimals: number;
+  private _volumeDecimals: number;
```
haVincy, line 49: I feel we can wrap these in a object param to reduce the growing customized params in CandyShop class.

```ts
export type CurrencyOption = {
  symbol : string;
  decimals: number;
  baseUnits: number;
  priceDecimals: number;
  volumeDecimals: number;
}

private _currencyOption: CurrencyOption;
```

lib/src/core/CandyShop/index.tsx
```diff
@@ -42,13 +42,22 @@ export class CandyShop {
...
   constructor(
     candyShopCreatorAddress: web3.PublicKey,
     treasuryMint: web3.PublicKey,
     candyShopProgramId: web3.PublicKey,
-    env: web3.Cluster
+    env: web3.Cluster,
+    currencySymbol: string,
+    currencyDecimals: number,
+    priceDecimals: number,
+    volumeDecimals: number
```
haVincy, line 60: So in constructor param, we can use CurrencyOption type.

```ts
constructor (
    candyShopCreatorAddress: web3.PublicKey,
    treasuryMint: web3.PublicKey,
    candyShopProgramId: web3.PublicKey,
    env: web3.Cluster,
    currencyOption: CurrencyOption
) {
   ...
   thus,_currencyOption = currencyOption;
}
```

lib/src/core/CandyShop/index.tsx
```diff
@@ -111,6 +126,22 @@ export class CandyShop {
...
+  }
+
+  get priceDecimals(): number {
+    return this._priceDecimals;
+  }
+
+  get volumeDecimals(): number {
+    return this._volumeDecimals;
+  }
+
```
haVincy, line 144: Not sure the use cases but I think we can simply return the currencyOption object.

```ts
get currencyOption(): CurrencyOption {
   return this._currencyOption;
}
```

lib/src/core/CandyShop/index.tsx
beepsaidblue: will refactor into settings param which is optional

lib/src/core/Orders/index.tsx
```diff
@@ -48,7 +48,6 @@ interface OrdersProps {
   walletConnectComponent: React.ReactElement;
   wallet: AnchorWallet | undefined;
   url?: string;
-  filterIdentifiers?: Array<string>;
```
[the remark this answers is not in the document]
beepsaidblue: this included in the filter PR

lib/src/core/CandyShop/index.tsx
beepsaidblue: ok

example/CandyShopContent.tsx
beepsaidblue: will remove

lib/src/components/BuyModal/BuyModalDetail.tsx
beepsaidblue: Will add try catch block

lib/src/core/CandyShop/index.tsx
beepsaidblue: okay

lib/src/core/CandyShop/index.tsx
beepsaidblue: refactored into settings

lib/src/components/CancelModal/CancelModalDetail.tsx
beepsaidblue: okay

lib/src/core/CandyShop/index.tsx
beepsaidblue: kept as is as easier to call the exact param when its needed

haVincy: kk.

README.md
```diff
@@ -51,9 +51,17 @@ const candyShop = new CandyShop(
   new PublicKey("csa8JpYfKSZajP7JzxnJipUL3qagub1z29hLvp578iN"), // Candy Shop program id
   "devnet", // mainnet, devnet
   wallet! // user wallet address
+  settings // (optional) additional shop settings
```
haVincy, line 54: wallet can be removed now?

lib/src/core/CandyShop/index.tsx
```diff
@@ -42,14 +54,24 @@ export class CandyShop {
...
-    env: web3.Cluster
+    env: web3.Cluster,
+    settings?: {
+      currencySymbol?: string;
+      currencyDecimals?: number;
+      priceDecimals?: number;
+      volumeDecimals?: number;
+    }
   ) {
+    settings = settings ?? {};
```
haVincy, line 73: We don't this, it will be assigned on line 84, if there's no `settings`, the check `settings.?.xxx` will be falsy and return the right hand value which is  DEFAULT params.

PR #75, opened 2022-04-15
chore: singleton instance, add react lint

vanminh1701:


lib/src/core/Sell/index.tsx
```diff
@@ -71,29 +72,27 @@ export const Sell: React.FC<SellProps> = ({
...
     if (!connection || !walletPublicKey || !candyShop) return;
 
-    if (loadingStatus === LoadStatus.ToLoad) {
-      fetchWalletNFTs(walletPublicKey, connection);
-      fetchOrders(walletPublicKey);
-    }
-  }, [connection, walletPublicKey, candyShop]);
+    fetchWalletNFTs(walletPublicKey, connection);
+    fetchOrders(walletPublicKey);
+  }, [connection, walletPublicKey, candyShop, fetchWalletNFTs, fetchOrders]);
```
haVincy, line 95: Have you tested the `fetchWalletNFTs` will not called twice after this modifications? Cuz if any `setState`, the `useEffect` will be called again. I still think we should add the if (loadingStatus === LoadStatus.ToLoad) to make sure not fetch NFT metadata concurrently. 

Even the react hook will remind you to add `loadingStatus` in dep array, it will not call again while loadingStatus being update cuz we will only fetch it when `ToLoad`, and we can control to refetch data by setLoadingStatus as ToLoad in future.

lib/src/core/OrderDetail/index.tsx
```diff
@@ -157,7 +157,7 @@ export const OrderDetail: React.FC<OrderDetailProps> = ({
             attributes={nftInfo?.attributes}
           />
 
-          {!wallet ? (
+          {!wallet?.publicKey ? (
```
beepsaidblue, line 160: why do we need reference to public key?

lib/src/core/Sell/index.tsx
```diff
@@ -71,29 +72,27 @@ export const Sell: React.FC<SellProps> = ({
...
       });
     },
-    [walletPublicKey]
+    [candyShopAddress]
   );
 
   useEffect(() => {
     if (!connection || !walletPublicKey || !candyShop) return;
 
-    if (loadingStatus === LoadStatus.ToLoad) {
```
beepsaidblue, line 92: how come we remove this?

lib/src/core/Sell/index.tsx
vanminh1701: I've tested, with 1 candyShop instance, it fetch. 1 time. Will we have a case when walletPublicKey change, and we need to fetch new list? If yes, the current LoadStatus will prevent that case.

haVincy: kk, I checkout the PR and tested. So far so good. Approved.

haVincy:
For future observation and debugging, could you add the console.log before those API calls in `core/OrderDetail` , `core/Orders`, `core/Sell`, and `core/Stat`. 

e.g.

```ts
const fetchOrders = useCallback(
    (walletPublicKey) => {
      console.log('core/Sell: fetchOrders');
      fetchOrdersByStoreIdAndWalletAddress(
        candyShopAddress,
        walletPublicKey.toString()
      ).then((sellOrders) => {
        setSellOrders(sellOrders);
      });
    },
    [candyShopAddress]
  );
```
```ts
 useEffect(() => {
     console.log('core/Stat: calling candyShop.stats()');
    candyShop
      .stats()
      .then((data: any) => {
        if (!data) return;
        setStat(data);
      })
      .catch((err) => {
        console.info('fetchOrdersByStoreId failed: ', err);
      });
  }, [candyShop]);
```
```
```

lib/src/core/OrderDetail/index.tsx
vanminh1701: I think I tested wrong case. Will not update this code

lib/src/core/Sell/index.tsx
beepsaidblue: I found a case where when user switches currencies in a multi-currency shop (they do so by updating candy shop instance), we want to re-call the fetchOrders method in the new currency, but we can keep the fetchWalletNfts call. is there a way we can implement this?

please look at the implementation here for multi currency sell:
https://github.com/LIQNFT/candy-machine-v2-with-marketplace

haVincy: > I found a case where when user switches currencies in a multi-currency shop (they do so by updating candy shop instance), we want to re-call the fetchOrders method in the new currency, but we can keep the fetchWalletNfts call. is there a way we can implement this?

Yes, we can separate `fetchOrders` and `fetchWalletNfts` by using different `setState` for them. e.g. `orderState` and `walletNftState`.

When we want to re=call the `fetchOrders`, we can set `orderState` as `ToLoad`.

RustySol (approved):

haVincy (approved):
LGTM.

PR #77, opened 2022-04-15
[chore] Update readme and add documentation ts comment

haVincy:


README.md
```diff
@@ -63,23 +63,35 @@ const candyShop = new CandyShop(
...
 
-```
+```ts
 import { Orders } from '@liqnft/candy-shop';
-
-<Orders
-  wallet={wallet}
-  candyShop={candyShop}
-  walletConnectComponent={<WalletMultiButton />}
-/>
```
haVincy, line 82: Remove details of these cuz we seems will keep updating when needed, just put the import usage here.

lib/src/index.tsx
```diff
@@ -1,6 +1,6 @@
 import './index.less';
 
-export { CandyShop } from './core/CandyShop';
+export { CandyShop, CandyShopSettings } from './core/CandyShop';
```
haVincy, line 3: Export the type so others can use.

PR #79, opened 2022-04-17
Add common package

RustySol:
Common package should contain all the API/program
call related logic, to be able to support cli in
the coming changes.


common/src/api/backend/OrderAPI.ts
```diff
@@ -14,9 +14,10 @@ export type OrdersFilterQuery = {
 };
 
 export async function fetchOrdersByStoreId(
+  axiosInstance: AxiosInstance,
   storeId: string,
   ordersFilterQuery: OrdersFilterQuery,
-  identifiers?: number[]
+  identifiers?: number[],
```
haVincy, line 20: nit - additional comma?

common/src/api/backend/NftAPI.ts
```diff
@@ -1,7 +1,10 @@
 import { SingleBase, Nft } from 'solana-candy-shop-schema/dist';
-import axiosInstance from '../../config/axiosInstance';
+import { AxiosInstance } from 'axios';
```
haVincy, line 2: What's difference between previous axio usage and current one, seems this one is from `axio` library itself?

lib/src/core/Sell/index.tsx
```diff
@@ -14,7 +12,11 @@ import {
   Order as OrderSchema,
   WhitelistNft,
 } from 'solana-candy-shop-schema/dist';
-import { CandyShop } from 'core/CandyShop';
+import {
+  CandyShop,
+  fetchNftsFromWallet,
+  SingleTokenInfo,
```
haVincy, line 18: nit - additional comma?

lib/src/index.tsx
```diff
@@ -1,9 +1,8 @@
 import './index.less';
 
-export { CandyShop, CandyShopSettings } from './core/CandyShop';
+export * from '@liqnft/candy-shop-common';
```
haVincy, line 3: I think we should specify the import here? Any reason that must using `*`?

common/src/api/backend/OrderAPI.ts
RustySol: this is from the formatter, somehow wasnt formatted before

common/src/api/backend/NftAPI.ts
RustySol: there is no difference, adding this is just for importing the typing
and make sure when the candyShop calls these methods, the axiosInstance is passed in from candyShop class

common/src/api/backend/OrderAPI.ts
haVincy: kk, I think I know why, the prettierrc should add ` trailingComma: 'none'`, I will take care this in my PR.

haVincy (approved):
LGTM.

PR #80, opened 2022-04-17
Add cli

RustySol:


sdk/tsconfig.json
```diff
@@ -1,27 +1,101 @@
...
+    /* Projects */
+    // "incremental": true,                              /* Enable incremental compilation */
+    // "composite": true,                                /* Enable constraints that allow a TypeScript project to be used with project references. */
+    // "tsBuildInfoFile": "./",                          /* Specify the folder for .tsbuildinfo incremental compilation files. */
+    // "disableSourceOfProjectReferenceRedirect": true,  /* Disable preferring source files instead of declaration files when referencing composite projects */
+    // "disableSolutionSearching": true,                 /* Opt a project out of multi-project reference checking when editing. */
+    // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
+
+    /* Language and Environment */
+    "target": "es2016",                                  /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
```
haVincy, line 14: Why change to use `es2016` instead of `es2019` ?
BTW, could we remove those comments, feel a bit cluttered,

cli/README.md
```diff
@@ -0,0 +1,23 @@
+CandyShop CLI
```
haVincy, line 1: We can also put `--help`  in readme? It seems commander.js support it by default. 

```ts
ts-node src/cli.ts --help
```

haVincy (approved):
I have not tested it.. will help to test. Let's merge first if it's not affecting others.

PR #81, opened 2022-04-17
Fix Tx size too large

RustySol:


RustySol:
Need to bump sdk version become merging

RustySol:
replace with #82

PR #85, opened 2022-04-18
listed for sell in Orders

vanminh1701:


lib/src/components/Order/index.tsx
```diff
@@ -34,10 +38,10 @@ export const Order: React.FC<OrderProps> = ({
     } catch (err) {
       return null;
     }
-  }, [order]);
+  }, [candyShop.baseUnitsPerCurrency, candyShop.priceDecimals, order?.price]);
 
   const onClose = useCallback(() => {
-    setSelection(null);
+    setSelection(undefined);
   }, []);
```
haVincy, line 45: Do we really need `useCallback` in this case?

lib/src/core/Orders/index.tsx
```diff
@@ -73,7 +73,7 @@ export const Orders: React.FC<OrdersProps> = ({
     number[] | undefined
   >(undefined);
 
-  const loadNextPage = (startIndex: number, limit: number) => {
+  const loadNextPage = (startIndex: number, limit: number) => () => {
```
haVincy, line 76: Why we need second arrow function here..?

lib/src/core/Orders/index.tsx
```diff
@@ -179,20 +178,18 @@ export const Orders: React.FC<OrdersProps> = ({
...
                   orders={orders}
                   walletConnectComponent={walletConnectComponent}
                   wallet={wallet}
                   candyShop={candyShop}
                   url={url}
                   hasNextPage={hasNextPage}
-                  loadNextPage={() =>
-                    loadNextPage(startIndex, ORDER_FETCH_LIMIT)
-                  }
+                  loadNextPage={loadNextPage(startIndex, ORDER_FETCH_LIMIT)}
```
haVincy, line 189: It seems should be following...?
```ts
loadNextPage={(startIndex, ORDER_FETCH_LIMIT) => loadNextPage(startIndex, ORDER_FETCH_LIMIT)}
```

lib/src/components/Order/index.tsx
vanminh1701: no, don't need it

lib/src/core/Orders/index.tsx
vanminh1701: startIndex, ORDER_FETCH_LIMIT come from current component, thay aren't arguments from `next` function in child componenet (InfiniteScroll)

PR #89, opened 2022-04-19
Update dependency

RustySol:


lib/package.json
```diff
@@ -50,7 +50,7 @@
     "typescript": "^4.1.3"
   },
   "dependencies": {
-    "@liqnft/candy-shop-sdk": "0.4.7",
+    "@liqnft/candy-shop-sdk": "0.4.9",
```
haVincy, line 53: Does this mean we will need to maintain the sdk with fixed version in `lib` as well?

lib/src/core/Orders/index.tsx
```diff
@@ -179,20 +178,18 @@ export const Orders: React.FC<OrdersProps> = ({
                       </FlexItem>
                     ))}
                 </Flex>
-              ) : !loading && !orders.length ? (
-                <Empty description="No orders found" />
-              ) : (
+              ) : !orders.length ? (
```
haVincy, line 181: This should be `orders.length ?`.

lib/package.json
RustySol: ideally yes, unless sdk has some major upgrade

lib/src/core/Orders/index.tsx
RustySol: let me rebase the commit, I think the history looks dirty

RustySol:
Closing for #91

PR #98, opened 2022-04-20
Feat/update UI sell modal

vanminh1701:


core/ui/src/components/BuyModal/BuyModalDetail.tsx
```diff
@@ -52,6 +41,19 @@ const BuyModalDetail: React.FC<BuyModalDetailProps> = ({
...
+    try {
+      return (
+        Number(order?.price) / candyShop.baseUnitsPerCurrency
+      ).toLocaleString(undefined, {
+        minimumFractionDigits: candyShop.priceDecimals,
+        maximumFractionDigits: candyShop.priceDecimals
+      });
+    } catch (err) {
+      return null;
+    }
```
haVincy, line 54: Why this need try/catch block, what kind of error it want to avoid?

core/ui/src/components/SellModal/index.tsx
```diff
@@ -92,7 +108,27 @@ export const SellModal: React.FC<SellModalProps> = ({
...
+  useEffect(() => {
+    //prettier-ignore
+    console.log('getTokenMetadataByMintAddress', { mint: nft.tokenMintAddress });
+    setLoading(true);
+    getTokenMetadataByMintAddress(nft.tokenMintAddress, connection)
+      .then((data: NftMetadata) => {
+        setToken(data);
+      })
+      .catch((err) => {
+        console.log('ERROR get getTokenMetadataByMintAddress', err);
```
haVincy, line 120: redundant 'get' ?

core/ui/src/components/SellModal/index.tsx
```diff
@@ -92,7 +108,27 @@ export const SellModal: React.FC<SellModalProps> = ({
...
+      })
+      .finally(() => {
+        setLoading(false);
+      });
+  }, [connection, nft.tokenMintAddress]);
+
   const isSubmit = formState.price !== undefined;
```
haVincy, line 127: Should we rename this to `isValidPrice` ?

core/ui/src/components/SellModal/index.tsx
```diff
@@ -130,15 +170,38 @@ export const SellModal: React.FC<SellModalProps> = ({
                 <span>{candyShop.currencySymbol}</span>
               </div>
 
-              <div className="candy-sell-modal-transaction">
+              <div className="candy-sell-modal-transaction candy-sell-modal-transaction--mb0">
```
haVincy, line 173: what is this `candy-sell-modal-transaction--mb0` ..? `className` should not contain consecutive `-`.

core/ui/src/public/Sell/index.tsx
```diff
@@ -50,7 +52,28 @@ export const Sell: React.FC<SellProps> = ({
...
 
+  // get fee
+  useEffect(() => {
+    if (!candyShop || !walletPublicKey) return;
+    setShopLoading(LoadStatus.Loading);
+    candyShop
+      .fetchShopByWalletAddress()
+      .then(({ result }: ListBase<CandyShopResponse>) => {
+        //prettier-ignore
+        setShop(result.find((item) => item.candyShopAddress === candyShop.candyShopAddress.toString()));
```
haVincy, line 66: Why adding prettier-ignore in this line?

core/ui/src/public/Sell/index.tsx
```diff
@@ -118,11 +141,15 @@ export const Sell: React.FC<SellProps> = ({
     );
   }
 
+  const loading =
+    loadingStatus !== LoadStatus.Loaded ||
+    orderLoading === LoadStatus.Loading ||
```
haVincy, line 146: I think we can let `orderLoading` to align others. `orderLoading !== LoadStatus.Loaded`.

core/ui/src/components/BuyModal/BuyModalDetail.tsx
vanminh1701: it's not my code.

core/ui/src/components/SellModal/index.tsx
vanminh1701: I follow style like BEM, B__E--M to know it is a modifier

core/ui/src/components/SellModal/index.tsx
vanminh1701: okay

core/ui/src/components/SellModal/index.tsx
vanminh1701: 👍

core/ui/src/public/Sell/index.tsx
vanminh1701: why can't i ignore format of prettier for this line? The purpose is for better view, nothing more.

core/ui/src/components/BuyModal/BuyModalDetail.tsx
beepsaidblue: to handle edge case where lines 46-51 returns an error, I think this case should be very remote

haVincy: 46-51 returns error can only happen Number(undefined) , so it seems what we need to do here is checking `order` before doing `Number(order.price) / ...`.

core/ui/src/public/Sell/index.tsx
haVincy: Then we should extend the `printWidth` in `.prettierrc.json` to 120 rather than adding ignoring?

vanminh1701: no, I think it's a bad idea. Cuz it will format lots of files. and I use this for a some lines where we don't need to focus more about the function find than the action purpose of that code is to setShop.

haVincy: hmm,, okay, I will find sometime to update prettier to avoid this ignore adding as possible.

haVincy (approved):

beepsaidblue (approved):

core/sdk/src/CandyShop.ts
```diff
@@ -396,4 +401,13 @@ export class CandyShop {
     });
     return fetchOrderByTokenMint(axiosInstance, mintAddress);
   }
+
+  public async fetchShopByWalletAddress(): Promise<
```
RustySol, line 405: should not use this API, not efficient, we need to make another api for query Shop by Id not walletAddress

core/ui/src/public/Sell/index.tsx
```diff
@@ -51,9 +53,35 @@ export const Sell: React.FC<SellProps> = ({
...
+  const [shop, setShop] = useState<CandyShopResponse>();
 
   const { refetch } = useContext(CandyContext);
 
+  // get fee
+  useEffect(() => {
+    if (!candyShop || !walletPublicKey) return;
+    setShopLoading(LoadStatus.Loading);
+    candyShop
+      .fetchShopByWalletAddress()
```
RustySol, line 66: also should use the getShop info backend API, still pending

RustySol:
@haVincy  @beepsaidblue if all good I think we can merge first. The API change can come in later releases

RustySol (approved):

core/ui/src/components/SellModal/index.tsx
```diff
@@ -1,37 +1,55 @@
...
   const [formState, setFormState] = useState<{ price: number | undefined }>({
     price: undefined
   });
   const [state, setState] = useState(TransactionState.DISPLAY);
-
-  const regex3Decimals = new RegExp('^[0-9]{1,11}(?:.[0-9]{1,3})?$');
+  const [token, setToken] = useState<NftMetadata>();
+  const [loading, setLoading] = useState<boolean>(true);
+  const [royalties, setRoyalties] = useState<number>();
+  const [transactionFee, setTransactionFee] = useState<number>(
```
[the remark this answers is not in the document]
RustySol: Yes, it should not be used, not sure why we did not exclude it.
Sell modal will only display tx Fee related data, won't set anything

RustySol (approved):

PR #103, opened 2022-04-21
Fix tx size too large

RustySol:


RustySol:
for nfts having many creators (>= 4 based on observation) we need to split buy and executeSales into 2 transactions, otherwise will get error like `Error: Transaction too large: 1267 > 1232`

RustySol:
can only merge after the indexer update

RustySol:
managed to bring it down to 
```Error: Transaction too large: 1255 > 1232```
Still need more investigation

RustySol:
ref: https://github.com/solana-labs/solana/issues/26391 possible this will solve the issue

pbredenberg:
> Documented in gitbook https://liqnft.gitbook.io/docs/candy-shop/marketplace
> 
> ref: [solana-labs/solana#26391](https://github.com/solana-labs/solana/issues/26391) possible this will solve the issue

I appreciate the followup!

RustySol:
> @RustySol @pbredenberg If this transaction size limit issue only happens for 4 or 5 creators, why can't we use Fanout wallet from metaplex? https://docs.metaplex.com/programs/hydra/intro You can just use 1 fanout wallet as a creator, which it has more than 4 members. Members can claim for their rewards from the Fanout wallet.

Hey thanks for the response. The thing is that the NFT is typically not created from us, so we have no say on which wallet to put it into the `creators` array. We are still working on the tx v2 feature in support of this.

RustySol:
Waiting on Anchor to support tx v2 https://github.com/coral-xyz/anchor/issues/2220

mPaella:
you can use lookup tables to reduce the tx size

PR #121, opened 2022-04-26
Add CHANGELOG

RustySol:


RustySol:
> I think let's just use tag's notes to update change log in each release? It seems to maintain tag's notes & this global changelog is time consuming for us now.
> 
> * https://github.com/facebook/react/releases

ok Sounds good

PR #133, opened 2022-05-01
[sdk] Refactor sdk and add documentation to CandyShop

nibbus0x:


core/sdk/src/api/utils.ts
```diff
@@ -12,29 +12,60 @@ import {
...
+  metadata: web3.PublicKey;
+}
+
+export interface SellTransactionArgs extends CandyShopTransactionArgs {
+  metadata: web3.PublicKey;
+}
+
+export interface CancelTransactionArgs extends CandyShopTransactionArgs {
+  tradeState: web3.PublicKey;
+}
```
haVincy, line 51: Looks better! Could we create a `model.ts` under `api` and move these interfaces into it.

core/sdk/src/CandyShop.ts
```diff
@@ -153,50 +211,68 @@ export class CandyShop {
       throw new Error('Idl not found');
     }
   }
-
+  /**
+   * Get treasury mint for this Candy Shop
+   */
```
haVincy, line 216: I feel we don't have to add the comments above each getters, if the method name is descriptive enough. And it's the getter for private members which is set up from constructor that already has @param notations.

core/sdk/src/CandyShop.ts
```diff
@@ -50,19 +50,68 @@ const DEFAULT_MAINNET_CONNECTION_URL = 'https://ssc-dao.genesysgo.net/';
...
+
+/**
+ * Arguments required for calling Candy Shop sell action
+ */
+export interface CandyShopSellArgs extends CandyShopActionArgs {}
+
+/**
+ * Arguments required for calling Candy Shop cancel action
+ */
+export interface CandyShopCancelArgs extends CandyShopActionArgs {}
```
haVincy, line 114: Can we create a `CandyShopModel.ts` under `sdk/src` to gather these interfaces rather than put it together along with growing CandyShop class object.

core/sdk/src/CandyShop.ts
```diff
@@ -291,30 +368,31 @@ export class CandyShop {
...
       program
-    );
+    });
     return txHash;
   }
+  /**
+   * Executes Candy Shop __Cancel__ action
+   *
+   * @param {CandyShopCancelArgs} args required arguments for cancel action
+   */
```
haVincy, line 392: Would like to know why have to use `__` ?

core/sdk/src/api/program/InsBuyAndExecuteSale.ts
```diff
@@ -368,16 +373,16 @@ async function sendTx(
...
-        })();
+  const signedTx = await ('signTransaction' in wallet
+    ? wallet.signTransaction(transaction)
+    : (async () => {
+        transaction.sign({
+          publicKey: wallet.publicKey,
+          secretKey: wallet.secretKey
+        });
+        return transaction;
+      })());
```
haVincy, line 384: Can we spilt this out as private method to get the `singedTx` 

```ts
private getSignedTx = async(wallet: AnchorWallet | web3.Keypair, transaction: // not sure type) => {
   if (wallet.signTransaction) {
      return wallet.signTransaction(transaction);
   } 
    transaction.sign({
          publicKey: wallet.publicKey,
          secretKey: wallet.secretKey
        });
      return transaction;
}
export async function buyAndExecuteSale(...) {
  ...
  const signedTx = await getSignedTx(wallet, transaction);
}


```

core/sdk/src/api/program/buyAndExecuteSale.ts
```diff
@@ -321,16 +326,16 @@ async function sendTx(
...
+  const signedTx = await ('signTransaction' in wallet
+    ? wallet.signTransaction(transaction)
+    : (async () => {
+        transaction.sign({
+          publicKey: wallet.publicKey,
+          secretKey: wallet.secretKey
+        });
+        return transaction;
+      })());
+
```
haVincy, line 338: Is it complete same from the `InsBuyAndExecuteSale` ..? If yes, we can make the `getSignedTx` as public to be reused for both.

core/sdk/src/CandyShop.ts
nibbus0x: That's markdown syntax. The surrounded words will display as bold in the HTML typedoc.

core/sdk/src/api/program/InsBuyAndExecuteSale.ts
nibbus0x: `if(wallet.signTransaction)` causes a Tyepescript error because `signTransaction` is not a method on the `web3.Keypair` type.

haVincy: oh... okay, then using `in`

core/sdk/src/CandyShopModel.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Arguments required for calling Candy Shop sell action
+ */
+export interface CandyShopSellArgs extends CandyShopActionArgs {}
+
+/**
+ * Arguments required for calling Candy Shop cancel action
+ */
+export interface CandyShopCancelArgs extends CandyShopActionArgs {}
```
haVincy, line 66: We also need to update the `export` in sdk/src/index.ts.

core/sdk/src/CandyShop.ts
haVincy: kk, new knowledge, I only know `__` is for the `python` private members/methods.

core/sdk/src/api/model.ts
```diff
@@ -0,0 +1,31 @@
+import { web3, BN, Program, Idl } from '@project-serum/anchor';
+import { AnchorWallet } from '@solana/wallet-adapter-react';
+
+export interface CandyShopTransactionArgs {
```
haVincy, line 4: nit - should call Params instead Args, param is the parameter with type that function will pass with, and the argument is the actual value that we pass to the function.

e.g. 
```ts
const testFunction = (param1: string, param2: number) => {
    console.log(param1);
    console.log(param2);
}
// '1' and 1 are the argument with actual value.
testFunction('1', 1);
```

haVincy (approved):
Left the minor comment, not affecting the overall PR.

core/sdk/src/api/model.ts
nibbus0x: Changed

core/sdk/src/CandyShopModel.ts
```diff
@@ -0,0 +1,66 @@
...
+  /** User wallet keypair */
+  wallet: AnchorWallet | web3.Keypair;
+}
+
+/**
+ * Arguments required for calling Candy Shop buy action
+ *
+ * @property {PublicKey} seller
+ */
+export interface CandyShopBuyParams extends CandyShopActionParams {
```
RustySol, line 53: nice one!

RustySol (approved):
LGTM

haVincy (approved):

PR #147, opened 2022-05-06
[sdk] Add ix to revoke seller treasurty ata delegate before sell tx (if exists)

nibbus0x:
[LIQ-526](https://liqnft.atlassian.net/browse/LIQ-526)

### DIFF
--- a/core/sdk/src/api/program/sell.ts
+++ b/core/sdk/src/api/program/sell.ts
@@ -1,8 +1,15 @@
 import * as anchor from '@project-serum/anchor';
-import { TOKEN_PROGRAM_ID } from '@solana/spl-token';
+import { createRevokeInstruction, getAccount, TOKEN_PROGRAM_ID } from '@solana/spl-token';
 import { SellTransactionParams } from '../model';
 import { AUCTION_HOUSE_PROGRAM_ID } from '../constants';
-import { getAuctionHouseProgramAsSigner, getAuctionHouseTradeState } from '../utils';
+import {
+  getAuctionHouseProgramAsSigner,
+  getAuctionHouseTradeState,
+  getAtaForMint,
+  sendTx,
+  treasuryMintIsNative
+} from '../utils';
+import { safeAwait } from '../../utils';
 
 export async function sellNft(params: SellTransactionParams): Promise<string> {
   const {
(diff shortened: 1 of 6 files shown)


core/sdk/src/utils/promiseUtils.ts
```diff
@@ -7,7 +7,7 @@ export function sleepPromise(ms: number): Promise<boolean> {
 }
 
 /* Helper buddy for removing async/await try/catch litter */
-export const safeAwait = (promise: Promise<any>, finallyCallback?: any) => {
+export function safeAwait<T>(promise: Promise<T>, finallyCallback?: any) {
```
haVincy, line 10: Any reason that we need generic type in safeAwait? Or just a minor refactor on this?

core/sdk/src/api/program/sell.ts
```diff
@@ -42,7 +44,22 @@ export async function sellNft(params: SellTransactionParams): Promise<string> {
   );
   const [programAsSigner, programAsSignerBump] = await getAuctionHouseProgramAsSigner();
 
-  const txHash = await program.methods
+  const transaction = new anchor.web3.Transaction();
+
+  const isNative = treasuryMint.equals(WRAPPED_SOL_MINT);
```
haVincy, line 49: I found this line in many places now. Maybe we can just move the isNative check as method in utils.ts and share with caller, if we want to change the determination for `isNative`,  we only need to modify one place.

BTW, is there a more descriptive but short naming than isNative?
```ts
export const isNative = (treasuryMint: web3.PublicKey): boolean => {
  return treasuryMint.equals(WRAPPED_SOL_MINT);
} 
```

core/sdk/src/utils/promiseUtils.ts
nibbus0x: So that you keep intellisense on the returned result type.

haVincy: But the return result of this `safeAwait` is an object...?
```ts
{
    result: any
    error: any
}
```

haVincy: hmm... okay, I think we only want to make sure the generic type of result when promise has been resolved.

BTW, to keep arrow function can do following
```ts
export const safeAwait = <T>(promise: Promise<T>, finallyCallback?: any)  =>
```

core/sdk/test/e2eToken.spec.ts
```diff
@@ -85,4 +87,35 @@ describe('e2e token flow', function () {
     const txHash = await connection.sendRawTransaction(transaction.serialize(), { skipPreflight: true });
     console.log('return NFT txHash', txHash);
   });
+
+  it('revokes seller treasury ata delagate', async () => {
+    const sellerTreasuryAta = await getAssociatedTokenAddress(TREASURY_MINT, user1.publicKey);
+    await approve(connection, user1, sellerTreasuryAta, web3.Keypair.generate().publicKey, user1, 1);
+
```
RustySol, line 94: can we add a check here to make sure delegate is really set?

core/sdk/src/api/utils.ts
```diff
@@ -179,7 +179,7 @@ export const checkPaymentAccountBalance = async (
       console.log('checkPaymentAccountBalance: getTokenAccountBalance error= ', accountBalance.error);
       paymentAccountBalance = undefined;
     } else {
-      paymentAccountBalance = accountBalance.result;
+      paymentAccountBalance = accountBalance.result.value.uiAmount;
```
RustySol, line 182: was this previously a bug?

nibbus0x: Probably, but not sure. When I added the generic type to `safeAwait` it started showing as an error.

RustySol: Is this testable in e2e test? I think if this was a bug should be reflected in e2e test?
If not maybe we need to increase the e2e test coverage

nibbus0x: So it was a bug. It's not reflected in the test cus apparently, when u ask js if something that is not a number is less than a number it returns false. i.e. `{} < 10 = false`. So if statement condition fails and no error thrown.

RustySol: Noted. Let's solve other cases in a separate PRs

haVincy (approved):

PR #150, opened 2022-05-08
[refactor] Provide a gateway to call CandyShop backend APIs

haVincy:


haVincy:
Currently, some information based APIs from backend must need a
CandyShop instance to call with which is not necessary, it lack of
flexibility to be adopted in difference cases.

CandyShop instance should be a part of sdk and not a major character,
the jobs only belongs to it are the transaction methods.

Separate the backend APIs into CandyShopInfoAPI.ts and not modify
any existed public method to keep backward compatibility.

Draft doc to understand the purpose if have doubt or any further suggestion
- https://drive.google.com/file/d/1B8Vchq4C3RQeDmVPkMfAaJprauEbapk8/view?usp=sharing

core/sdk/src/CandyShop.ts
```diff
@@ -211,45 +213,24 @@ export class CandyShop {
...
+      authority: auctionHouseAuthority,
+      authorityBump,
+      auctionHouse,
+      feeAccount,
+      candyShop: this._candyShopAddress,
+      price,
+      amount: new BN(1),
+      program
+    };
+    const txHash = await getBuyTxHash(this._programId, buyTxHashParams);
```
RustySol, line 233: this function doesn't only get buyTxHash, it also execute the operation onchain. Suggest we rename it to `buyAndExecuteSales`

core/sdk/src/CandyShopInfoAPI.ts
```diff
@@ -0,0 +1,84 @@
+/**
```
RustySol, line 1: Nice refactor!

core/sdk/src/CandyShop.ts
haVincy: okay

RustySol (approved):

PR #157, opened 2022-05-11
[core/ui] Rendering Order list by matching the order with treasuryMint list

vuvu2259:


core/ui/src/hooks/useHashTable.ts
```diff
@@ -0,0 +1,53 @@
...
+  },
+  '56pdaHboK66cxRLkzkYVvFSAjfoNEETJUsrdmAYaTXMJ': {
+    symbol: 'TEST',
+    decimals: 9,
+    logoURI: 'https://haycafe.vn/wp-content/uploads/2022/03/Avatar-hai-1.jpg'
+  }
+};
+
+export function useHashTable(): HooksReturnValue {
+  const [hashTable, setHashTable] = useState<HashTable>(HASH_TABLE);
```
haVincy, line 35: Can we just try `Map` and rename this to `useShopExchangeInfo` to return the object `ShopeExchangeInfo`?

```ts
export interface ShopExchangeInfo {
    symbol: string;
    decimals: number;
    logoURI: string;
}

const [exhangeInfoMap, setExchangeInfoMap] = useState<Map<string, ShopExchangeInfo>>();


```

core/ui/src/components/Order/index.tsx
```diff
@@ -8,9 +8,11 @@ import { LiqImage } from 'components/LiqImage';
 import { CancelModal } from 'components/CancelModal';
 
 import { Order as OrderSchema } from 'solana-candy-shop-schema/dist';
-import { CandyShop } from '@liqnft/candy-shop-sdk';
+import { CandyShop, fetchShopByCreatorAddress } from '@liqnft/candy-shop-sdk';
 
 import './index.less';
+import { HashTable, useHashTable } from 'hooks/useHashTable';
+import { useEffect } from 'react';
```
vanminh1701, line 15: please move this to 1st line, we have already imported react

core/ui/src/components/Order/index.tsx
```diff
@@ -32,6 +35,23 @@ export const Order: React.FC<OrderProps> = ({ order, wallet, walletConnectCompon
   const [selection, setSelection] = useState<OrderSchema>();
   const [orderCandyShop, setOrderCandyShop] = useState<CandyShop>(candyShop);
 
+  const { hashTable, setHashTable } = useHashTable();
+
+  useEffect(() => {
```
vanminh1701, line 40: If we check treasuryMint for each Order component, it will have mulitiple requests to get the same treasuryMint information. I think we should request new treasuryMint after api candyShop.order() finish and before render UI

core/ui/src/hooks/useHashTable.ts
```diff
@@ -0,0 +1,53 @@
...
+  const [hashTableStringify] = useState<string | HashTableEnum>(
+    localStorage.getItem('hashTable') || HashTableEnum.EMPTY
+  );
+
+  useEffect(() => {
```
vanminh1701, line 40: this effect will be run 1 time cuz hashTableStringify won't be changed, so just create hashTableStringify inside useEffect, and we don't need hashTableStringify state anymore
```
useEffect(() => {
  const hashTableStringify = localStorage.getItem('hashTable') || HashTableEnum.EMPTY
  if (hashTableStringify === HashTableEnum.EMPTY) {
      localStorage.setItem('hashTable', JSON.stringify(HASH_TABLE));
    } else {
      setHashTable(JSON.parse(hashTableStringify));
    }
})

```

core/sdk/src/api/backend/ShopAPI.ts
```diff
@@ -14,6 +14,13 @@ export async function fetchShopByShopId(axiosInstance: AxiosInstance, shopId: st
   return axiosInstance.get<SingleBase<CandyShop>>(`/shop/id/${shopId}`).then((response) => response.data);
 }
 
+export async function fetchShopByCreatorId(
+  axiosInstance: AxiosInstance,
+  creatorId: string
+): Promise<SingleBase<CandyShop[]>> {
+  return axiosInstance.get<SingleBase<CandyShop[]>>(`/shop/${creatorId}`).then((response) => response.data);
+}
```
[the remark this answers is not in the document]
haVincy: Change name to `fetchShopsByCreatorId` and please add comment that this is restricted to certain creators.

haVincy:
As discussed in channel, move out the local storage logic in core/ui, let `Order` component to get the treasuryMint list info from the consumer (currently, only LIQNFT case) and render the UI for all orders if needed.

PR #158, opened 2022-05-13
[core/ui] display difference order in liqnft case

vuvu2259:


core/ui/src/components/Order/index.tsx
```diff
@@ -54,8 +49,25 @@ export const Order: React.FC<OrderProps> = ({ order, wallet, walletConnectCompon
     }
   };
 
+  const exchangeInfo = useMemo<ShopExchangeInfo>(() => {
+    if (!order.symbol || !order.decimals || !order.logoURI) {
+      return {
+        symbol: candyShop.currencySymbol,
```
vanminh1701, line 55: it should be a constant SOL object mint info in constant folder and import to  component. Cuz the logoURI is SOL but candyShop information maybe not a SOL, it will make a confusion 
And we don't need useMemo here, component still re-render with useMemo, just use a function outside component is enough

core/ui/src/utils/getPrice.ts
```diff
@@ -0,0 +1,12 @@
+import { CandyShop } from '@liqnft/candy-shop-sdk';
+import { ShopExchangeInfo } from 'model';
+import { Order as OrderSchema } from 'solana-candy-shop-schema/dist';
+
+export const getPrice: any = (candyShop: CandyShop, order: OrderSchema, exchangeInfo: ShopExchangeInfo) => {
+  if (!order?.price || !exchangeInfo) return null;
```
vanminh1701, line 6:
```
if (!order?.price || exchangeInfo?.decimals !== undefined) return null;
``` 
make sure we must have decimal from exchangeInfo

core/ui/src/components/Order/index.tsx
vuvu2259: "If order doesn't have those field, it will fallback to configure" => Need candyShop instance to get default settings
=> Cannot import an default object from constant file => Can import a function that need candyShop instance and order to cook it
Default settings don't have logoURI so what we need to fallback, already ask in discord

core/ui/src/utils/getPrice.ts
vuvu2259: always has decimals in `exchangeInfo`

vanminh1701: what if BE didn't return decimals?, we will have NAN value here

core/ui/src/model.ts
```diff
@@ -3,3 +3,8 @@ export enum TransactionState {
   PROCESSING,
   CONFIRMED
 }
+
+export interface ShopExchangeInfo {
+  symbol: string;
+  decimals: number;
+}
```
haVincy, line 10: Why we don't need logoURI anymore?

core/ui/src/utils/getPrice.ts
vanminh1701: got it

haVincy:
```
if (!order.price || !exchangeInfo?.decimals) return null;
```

core/ui/src/model.ts
haVincy: okay... I saw the conversation in chat.

core/ui/src/utils/getPrice.ts
haVincy: okay, resolved.

core/ui/src/utils/getExchangeInfo.ts
```diff
@@ -0,0 +1,17 @@
...
+      symbol: candyShop.currencySymbol,
+      decimals: candyShop.currencyDecimals
+    };
+  }
+
+  return {
+    symbol: order.symbol,
+    decimals: order.decimals
+  };
+}
```
haVincy, line 17: We can do following 

```ts
export function getExchangeInfo(order: OrderSchema, candyShop: CandyShop): ShopExchangeInfo {
 return {
    symbol: order.symbol ?? candyShop.currencySymbol,
    decimals: order.decimals ??  candyShop.currencyDecimals
  };
}
```

vanminh1701 (approved):

core/ui/src/utils/getExchangeInfo.ts
vuvu2259: already change it

haVincy:
Tested at local, build failed, don't merge until fix.

PR #163, opened 2022-05-14
[chore] publish script

beepsaidblue:
Strawman publish script to reduce publishing pain (NEEDS TESTING)

TODO: how to prevent publish when there are any compile warnings in yarn build

Example usage:
1. Bump SDK and UI version and publish SDK and UI `bash publish.sh sdk=patch ui=patch`
2. Bump UI version and publish UI `bash publish.sh ui=patch`
3. specify which component to publish (sdk, ui)
4. specify which version to bump (patch, minor, major)

### DIFF
--- a/bump.ts
+++ b/bump.ts
@@ -1,0 +1,60 @@
+import fs from 'fs';
+
+/**
+ * Increments version of core/sdk and core/ui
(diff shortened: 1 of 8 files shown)


haVincy:
Can consider to use the CI from Marketplace 
- https://github.com/marketplace/actions/npm-publish

PR #167, opened 2022-05-14
Add Auction Types

RustySol:


core/types/src/response/auction.ts
```diff
@@ -0,0 +1,20 @@
+export enum AuctionStatus {
+  CREATED,
+  STARTED,
+  COMPLETE,
+  CANCELLED
+}
```
haVincy, line 6: Should we specify enum string to help debug.
```ts
export enum AuctionStatus {
  CREATED = 'CREATED'.
  STARTED = 'STARTED',
  COMPLETE =  'COMPLETE',
  CANCELLED = 'CANCELLED'
}
```

core/types/src/response/auctionBid.ts
```diff
@@ -0,0 +1,14 @@
+export enum BidStatus {
+  OPEN,
+  WITHDRAWN,
+  WON,
+  LOST
+}
```
haVincy, line 6: Same here

haVincy (approved):

core/types/src/response/auction.ts
RustySol: these are used as numbers

haVincy: You mean backend already define the numbers? Is it possible to change to use string enum for convenient debugging?

RustySol: yes the DB column is using number to indicate the statuses, just like `OrderStatus`

PR #180, opened 2022-05-22
[core] Support cache user NFT in IndexedDB

haVincy:


haVincy:
Note: once we have reached larger scale, will need service worker to do those massive async jobs in worker thread to avoid potential block for main thread in browser. We're cool to not have that now for candy-shop. So this IndexedDB wrap up only consider minimum use case.

beepsaidblue (approved):

PR #195, opened 2022-05-30
ui: update-Sell-component-status

vanminh1701:


haVincy (approved):

haVincy:
LGTM. Just make sure have tested, Backend increased the polling frequency from chain, the latency should be improved a bit.

haVincy:
Need to sort it out how to precisely reloading NFT by USER_NFT signal, currently, the updates from USER_NFT timestamp doesn't mean there's NFT changes in user's wallet.

PR #202, opened 2022-05-31
[LIQ-611] Make video NFTs playable in Candy Shop

vuvu2259:


core/ui/src/components/CancelModal/CancelModalDetail.tsx
```diff
@@ -51,10 +52,33 @@ export const CancelModalDetail: React.FC<CancelModalDetailProps> = ({
 
   const orderPrice = getPrice(candyShop, order, exchangeInfo);
 
+  const Preview = () => {
```
beepsaidblue, line 55: could we wrap this into it's own component called `Viewer` so it can be re-used by both buy and cancel modal. also need to add it to OrderDetail.

beepsaidblue (approved):

PR #205, opened 2022-06-01
[fix] remove import and fix typescript error

vuvu2259:


core/ui/src/declarations.d.ts
```diff
@@ -0,0 +1,5 @@
+declare namespace JSX {
+  interface IntrinsicElements {
+    'model-viewer': any;
+  }
+}
```
haVincy, line 5: I think we don't need a new file, we already have `.d.ts` under `core/ui/src`?

core/ui/src/public/OrderDetail/index.tsx
```diff
@@ -112,7 +111,13 @@ export const OrderDetail: React.FC<OrderDetailProps> = ({
...
     <div className="candy-order-detail">
       <div className="candy-container">
-        <div className="candy-order-detail-left">{order && <Viewer order={order} />}</div>
+        <div className="candy-order-detail-left">
+          {order && order.nftAnimationLink ? (
+            <video width="100%" height="auto" controls autoPlay loop src={order.nftAnimationLink}></video>
+          ) : (
+            <LiqImage src={order?.nftImageLink || ''} alt={order?.name} fit="contain" />
+          )}
+        </div>
```
haVincy, line 120: Why we have to change this if we can fix the error by declaration?

core/ui/src/declarations.d.ts
vuvu2259: I tried to add in `.d.ts` but it isn't work. You can try!

core/ui/src/public/OrderDetail/index.tsx
vuvu2259: @beepsaidblue said the order detail don't need to show 3d model so I change it in this branch as well
This isn't fix for the typescript error

core/ui/src/declarations.d.ts
haVincy: I tried but works for me?

```ts
declare namespace JSX {
  interface IntrinsicElements {
    "model-viewer": ModelViewerElement;
  }
}
```

haVincy (approved):

core/ui/src/public/OrderDetail/index.tsx
beepsaidblue: i meant OrderDetail don't need "import '@google/model-viewer". we should keep the Viewer here

beepsaidblue (approved):

PR #209, opened 2022-06-03
[core/ui] add more padding into auction card

vuvu2259:


core/ui/src/components/Modal/index.less
```diff
@@ -15,7 +15,8 @@
     left: 50%;
     transform: translate(-50%, -50%);
 
-    padding: 20px;
+    padding: 40px;
+    padding-bottom: 20px;
```
haVincy, line 19: Why don't we just
```less
padding: 40px 40px 20px 40px;
```

vuvu2259: It's not friendly for me :D

vuvu2259: I will update for u ^^

vuvu2259: updated

haVincy: If you prefer specific declaration, then shouldn't mix up with `padding` but 
```less
padding-top: 40px;
padding-bottom: 20px;
padding-left: 40px;
padding-right: 40px;
```

haVincy (approved):

PR #211, opened 2022-06-03
[fix] highest bidder should be able to withdraw after buy now

RustySol:


core/sdk/src/api/utils.ts
```diff
@@ -327,10 +327,15 @@ export const checkBidParams = async (auction: web3.PublicKey, bidPrice: BN, prog
...
+export const checkCanWithdraw = async (
+  auction: web3.PublicKey,
+  bid: web3.PublicKey,
+  program: Program,
+  isBuyNow = false
+) => {
   const auctionData = await getAuctionData(auction, program);
 
-  if (auctionData.highestBid && auctionData.highestBid.key.equals(bid)) {
+  if (!isBuyNow && auctionData.highestBid && auctionData.highestBid.key.equals(bid)) {
```
nibbus0x, line 338: Does `auctionData.highestBid` return a non-null value here? If so it may be because we are not waiting for buy now tx confirmation.

RustySol: Yes it's possible that `auctionData.highestBid` returns non-null, because this check is not only run in buyNow, itself is also a standalone operation on the UI.
More specifically, when this check is not run under buy now, i.e. user simply tries to withdraw Bid, `isBuyNow` will be `false`, and if the user is indeed the highest bidder, i.e. `auctionData.highestBid` is non null, then the check should not pass

nibbus0x: Sorry, I meant in the situation where this check is run after `BuyNow`, i.e. `isBuyNow = true`. In this case `BuyNow` should set `highestBid` to `null`, which would make the `!isBuyNow` check irrelevant since `auction.highestBid` is falsy.

So my question is, is `highestBid` ever non-null after calling `BuyNow`? If so, we're either not awaiting tx confirmation, or there's a problem with the program code.

RustySol: good point, in this case perhaps we dont need the `isBuyNow` check anymore

RustySol: I suspect the onchain status somehow does not propagate that fast, and therefore likely the `withdrawBid` call will still be rejected when the check is run (I tried on devnet)
Regardless, highest bidder even after buyout and still come back to press the `retrieve fund` button so overall I think is alright. Just right now need to communicate with those who run into error to try withdraw again.

RustySol:
Without `isBuyNow` this PR essentially has no new change. Closing

PR #206, opened 2022-06-02
[LIQ-626] Refactor sdk to support v2 CandyShop program

nibbus0x:

PR #218, opened 2022-06-07
Add CandyShop V2 support

RustySol:


RustySol:
close for https://github.com/LIQNFT/candy-shop/pull/206

PR #225, opened 2022-06-10
[LIQ-613] Remove withdrawBid after buyNow

RustySol:
Due to the state change propagation delay, there
are situations where higgest bidder fails to
withdrawBid after buyNow. Remove the logic for now
and highest bidder will need to go back and withdrawBid
after buyNow. Will need to come up with better UI
solution to it.


nibbus0x:
As an alternative solution, we could try to set up a system with (Cronos)[https://www.cronos.so/] that periodically checks bids that have are no longer winning and calls withdraw for them. I'll look more into the feasibility of that this weekend.

nibbus0x (approved):
Looks good!

PR #226, opened 2022-06-10
Update UI with latest sdk

RustySol:


core/ui/src/components/Order/index.tsx
```diff
@@ -35,13 +35,13 @@ export const Order: React.FC<OrderProps> = ({ order, wallet, walletConnectCompon
...
     } else {
       setOrderCandyShop(
-        new CandyShop(
-          new web3.PublicKey(order.candyShopCreatorAddress),
-          new web3.PublicKey(order.treasuryMint),
-          candyShop.programId,
-          candyShop.env,
-          candyShop.settings
-        )
+        new CandyShop({
```
nibbus0x, line 38: Is there a way to determine if this shop is an enterprise shop here? If so, can we pass the `isEnterprise` flag in the params obj. Otherwise it always defaults to false.

RustySol: The enterprise users usually use sdk directly

nibbus0x (approved):

core/ui/src/components/Order/index.tsx
nibbus0x: Ok true. Looks good then

PR #238, opened 2022-06-14
[LIQ-630] auction polling to update UI

vanminh1701:


core/sdk/src/api/backend/ShopAPI.ts
```diff
@@ -1,5 +1,11 @@
 import { SingleBase, WhitelistNft, ShopStatus, CandyShop, ListBase } from '@liqnft/candy-shop-types';
 import { AxiosInstance } from 'axios';
+import qs from 'qs';
+
+export interface ShopStatusShopQuery {
```
haVincy, line 5: We should move this to `core/types/query` and rename as `ShopStatusQuery`.

core/sdk/src/api/backend/ShopAPI.ts
```diff
@@ -16,9 +22,9 @@ export async function fetchShopByShopId(axiosInstance: AxiosInstance, shopId: st
...
   shopId: string,
-  walletAddress?: string
+  query: ShopStatusShopQuery
 ): Promise<SingleBase<ShopStatus[]>> {
-  return axiosInstance
-    .get<SingleBase<ShopStatus[]>>(`/status/${shopId}?${walletAddress ? `walletAddress=${walletAddress}` : ''}`)
-    .then((res) => res.data);
+  const { checkAuction, walletAddress } = query;
+  const queryString = qs.stringify({ walletAddress, checkAuction });
+  return axiosInstance.get<SingleBase<ShopStatus[]>>(`/status/${shopId}?${queryString}`).then((res) => res.data);
```
haVincy, line 29: Let's refactor a bit this `ShopAPI.ts`.

Make every API url call as const to make axios usage cleaner, e.g.
```ts
export async function fetchShopStatusByShopId(
    axiosInstance: AxiosInstance,
    shopId: string,
    query: ShopStatusQuery
): Promise<SingleBase<ShopStatus[]>> {
   const { checkAuction, walletAddress } = query;
   const queryString = qs.stringify({ walletAddress, checkAuction });
   const url = `/status/${shopId}?${queryString}`;
   return axiosInstance.get<SingleBase<ShopStatus[]>>(url).then((res) => res.data));
}
```

core/ui/src/public/Context/index.tsx
```diff
@@ -22,11 +22,13 @@ const Logger = 'CandyShopUI/CandyShopDataValidator';
...
   candyShopAddress?: web3.PublicKey;
   setCandyShopAddress: Dispatch<SetStateAction<web3.PublicKey | undefined>>;
+  setSubjects: Dispatch<SetStateAction<any>>;
 }
 
 export const CandyContext = createContext<ContextData>({
   candyShopAddress: undefined,
-  setCandyShopAddress: () => console.log(`${Logger}: update CandyShopAddress`)
+  setCandyShopAddress: () => console.log(`${Logger}: update CandyShopAddress`),
+  setSubjects: () => console.log(`${Logger}: update subject for polling`)
```
haVincy, line 31: Would it be helpful we log which subject is being updated?

haVincy (commented):
I believe this will need to wait BE deploy to staging (devnet) to merge?

core/sdk/src/api/backend/ShopAPI.ts
vanminh1701: okay

core/ui/src/public/Context/index.tsx
vanminh1701: it is default value for context provider, let me see if I can remove it. I added it cuz ts warning

core/ui/src/config/axiosInstance.ts
```diff
@@ -1,30 +0,0 @@
-import axios, { AxiosRequestConfig } from 'axios';
```
haVincy, line 1: Is it right that we delete this file...?

core/sdk/src/config/index.ts
```diff
@@ -22,6 +22,7 @@ export const configBaseUrl = (env: web3.Cluster): void => {
       // Do something before request is sent
       if (env === 'devnet') {
         config.baseURL = 'https://ckaho.liqnft.com/api/';
+        config.baseURL = 'http://localhost:8082/api/';
```
RustySol, line 25: please remove

core/types/src/query/query.ts
```diff
@@ -26,3 +26,8 @@ export interface ShopQs {
   offset?: number;
   limit?: number;
 }
+
+export interface ShopStatusQuery {
+  walletAddress?: string;
+  checkAuction?: boolean;
```
RustySol, line 32: maybe will be useful to comment here why we add a flag to specifically `checkAuction`? how about `Order` and `Trade`?

core/ui/src/constant/Orders.ts
```diff
@@ -1,6 +1,7 @@
 import { OrderSortBy } from '@liqnft/candy-shop-sdk';
 
 export const ORDER_FETCH_LIMIT = 12;
+export const BACKGROUND_UPDATE_LIMIT = ORDER_FETCH_LIMIT * 2;
```
RustySol, line 4: any reason why this is `ORDER_FETCH_LIMIT  * 2`?

core/sdk/src/config/index.ts
vanminh1701: oh, my bad

core/types/src/query/query.ts
vanminh1701: I see for each subject (order, trade, auction), BE need to query a table to check it. So what I want is FE will trigger a condition to check any subject. e.g we only check auction latest data when component auction is display on UI

core/ui/src/constant/Orders.ts
vanminh1701: I just think it will reduce number of requests to get list auctions. how about you?

core/sdk/src/config/index.ts
haVincy: We need this update to let test link works.

core/types/src/query/query.ts
haVincy: If we want to extend the same ShopStatus API, maybe we can be more flexible on this flag.
Can design an enum in types to for check flag on different data in future. 

```ts
export enum ShopStatusTargetType {
   Auction = 'Auction',
   // can put more in future.
}
```

core/ui/src/constant/Orders.ts
RustySol: all good

core/ui/src/public/Auction/Auctions.tsx
```diff
@@ -69,6 +81,45 @@ export const Auctions: React.FC<AuctionsProps> = ({ walletConnectComponent, wall
...
+    )
+      .then((responses: ListBase<Auction>[]) => {
+        const memo: { [key: string]: true } = {};
+        const newList = responses.reduce((acc: Auction[], res: ListBase<Auction>) => {
+          if (!res.success) return acc;
+          res.result.forEach((auction) => {
+            if (memo[auction.tokenAccount]) return;
+            memo[auction.tokenAccount] = true;
+            acc.push(auction);
+          });
```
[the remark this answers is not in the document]
vanminh1701:
```for (let i of undefined){
console.log(i)
}```
I tested this case on consolke, it through TypeError. How can for loop prevent it. 
I think performance here will not a big deal. same as the idea in your url conclusion: 
`Any difference in performance between the four looping mechanisms should normally not matter. If it does, you are probably doing something very computationally intensive and switching to WebAssembly may make sense.`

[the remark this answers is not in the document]
haVincy: hmm.. I see, I mistaken this yesterday night. Then can we add the `if` check before doing loop?

haVincy:
Test link - https://core-auction-polling.d22tqykss69onq.amplifyapp.com/

haVincy (approved):

PR #244, opened 2022-06-17
[sdk] Refactor buy method in CandyShop to have flexiblity

haVincy:


core/sdk/src/api/program/v2/marketplace/buyAndExecuteSale.ts
```diff
@@ -222,3 +223,48 @@ export async function buyAndExecuteSale(params: BuyAndExecuteSaleTransactionPara
...
+/**
+ * Chooses to call either v1 or v2 version of passed fuction based on candy shop version
+ *
+ * @param {any} params argument to the function to call
+ * @param {CandyShopVersion} version version of the candy shop
+ * @param {function} v1Func function to call if using v1 candy shop
+ * @param {function} v2Func function to call if using v1 candy shop
+ */
+// Please feel free to come up with better name :)
+export function call(
```
RustySol, line 259: `invokeCandyShop`?

core/sdk/src/CandyShopTrade.ts
```diff
@@ -0,0 +1,74 @@
...
+  getAuctionHouseTreasuryAcct,
+  getMetadataAccount,
+  getProgram
+} from './api';
+import { CandyShopTradeBuyParams } from './CandyShopModel';
+
+/**
+ * A abstract class to provide static trading methods without CandyShop instance but partial information from shop.
+ */
+export abstract class CandyShopTrade {
```
RustySol, line 19: Why couldn't we simply declare the `buy` function but instead encapsulate it in the `CandyShopTrade` abstract class?

core/sdk/src/api/program/v2/marketplace/buyAndExecuteSale.ts
haVincy: Does this is exposed already in publish? I think we might want to control what methods we can expose or what methods should not?

core/sdk/src/CandyShopTrade.ts
haVincy: I found it will be clear for callers when using our methods, it must use `CandyShopTrade.xxx` in codes and will have lesser possibility to mess up with other libraries.

core/sdk/src/api/program/v2/marketplace/buyAndExecuteSale.ts
haVincy: e.g. maybe we should not expose `src/api` in `core/sdk/index.ts`

core/sdk/src/api/utils.ts
```diff
@@ -24,9 +25,47 @@ import { AnchorWallet } from '@solana/wallet-adapter-react';
...
+    connection,
+    // check the instance type
+    wallet instanceof web3.Keypair ? getNodeWallet(wallet) : wallet,
+    options
+  );
+
+  const idl = candyShopProgramId.equals(CANDY_SHOP_V2_PROGRAM_ID) ? candyShopV2Idl : candyShopIdl;
+  // Directly use the JSON file here temporarily
+  // @ts-ignore
+  return new Program(idl, candyShopProgramId, provider);
```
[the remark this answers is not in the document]
haVincy: The reason it can be sync is we no longer use the methods that is returning Promise to get the `Program`. We're using the JSON file now, what I'm asking is do we plan to use the JSON file here from now on?
 
```ts
 const idl = candyShopProgramId.equals(CANDY_SHOP_V2_PROGRAM_ID) ? candyShopV2Idl : candyShopIdl;
```

core/sdk/src/api/program/v2/marketplace/buyAndExecuteSale.ts
haVincy: I will rename it in my followup `sdk/src/api` restructure PR.

RustySol (approved):

PR #249, opened 2022-06-21
[sdk] Refactor vendor/supplier to vendor/shipping and not expose it

haVincy:


RustySol (approved):

haVincy:
Would like to get @nibbus0x 's review before merging.

PR #253, opened 2022-06-22
[LIQ-654] custom URL for seller address

vanminh1701:


core/ui/src/public/OrderDetail/index.tsx
```diff
@@ -123,30 +128,14 @@ export const OrderDetail: React.FC<OrderDetailProps> = ({
...
-            </div>
-          </div>
+          {order?.tokenMint ? (
+            <NftStat
+              owner={order.walletAddress}
+              edition={order.edition}
+              tokenMint={order.tokenMint}
+              sellerUrl={sellerUrl}
+            />
+          ) : null}
```
haVincy, line 138: Do we need `null` here? Or we can just 
```ts
{ order?.tokenMint && <NftStat ... /> }
```

core/ui/src/utils/shortenAddress.ts
```diff
@@ -0,0 +1,5 @@
+const NUMBER_OF_CHAR = 4;
+
+export const shortenAddress = (address: string, chars = NUMBER_OF_CHAR): string => {
+  return `${address.slice(0, chars)}...${address.slice(-chars)}`;
+};
```
haVincy, line 5: Can we just put these in the `utils/format.ts` ?

core/ui/src/public/OrderDetail/index.tsx
vanminh1701: can do your code. And it will be the same result. 
just note a bad practice with `&&`: 
```
{0 && <Component />} // will render number 0 on UI
```

core/ui/src/utils/shortenAddress.ts
vanminh1701: okay, great idea

core/ui/src/public/OrderDetail/index.tsx
haVincy: hmm... then how should avoid it?

vanminh1701: yes we can do !!, but why we need more 2 operations instead of ternary.

haVincy (approved):

PR #276, opened 2022-07-04
fix: wrong baseURl when calling api from sdk

vanminh1701:


core/sdk/src/vendor/config.ts
```diff
@@ -4,7 +4,8 @@ import { web3 } from '@project-serum/anchor';
 const axiosInstance = axios.create({
   headers: {
     'Content-Type': 'application/json'
-  }
+  },
+  baseURL: 'https://ckaho.liqnft.com/api/'
```
haVincy, line 8: The API calls from SDK is using the exported one, if there's no `env` being setup, we can just do following and delete the code here?

```ts
const interceptorEvent = axiosInstance.interceptors.request.use(
    (config: AxiosRequestConfig<any>) => {
      // Default baseUrl is staging backend
      config.baseURL = 'https://ckaho.liqnft.com/api/';

      if (env === 'mainnet-beta') {
        config.baseURL = 'https://candy-shop.liqnft.com/api/';
      }
```

haVincy: Yup, not a strong comment, or maybe using current fix but add the comment to indicate that `baseURL` is BE's staging?

haVincy (approved):
Approved first. Minh, you can merge once update.

RustySol (approved):

PR #277, opened 2022-07-04
LIQ-667 error when create, cancel auction with programId v2

vanminh1701:


core/sdk/src/CandyShop.ts
```diff
@@ -409,7 +411,7 @@ export class CandyShop {
       program
     };
 
-    const txHash = await supply(createAuctionParams, this._version, createAuctionV1, this.createAuction);
```
haVincy, line 412: hmmm... I see, this seems the pitfalls of dup naming across files.... got to be careful. @RustySol @nibbus0x .

haVincy (approved):

core/sdk/src/CandyShop.ts
RustySol: noted

RustySol (approved):

PR #281, opened 2022-07-09
Remove request compute budget call

RustySol:
Due to the new change on compute budget (ref https://discord.com/channels/428295358100013066/749579745645166592/995070336586481866)
Screen Shot 2022-07-09 at 3 28 24 PM

We need to request extra compute budget when calling some auction related transactions


RustySol:
right now this feature is only enabled on devent, should merge and publish new candyShop version once its enabled on mainnet

PR #283, opened 2022-07-12
LIQ-689 logic allow sell whitelist nft

vanminh1701:


core/sdk/src/vendor/token/fetchNftsFromWallet.ts
```diff
@@ -132,7 +137,11 @@ const removeOutdatedNftFromIDB = async (
   // Remove nft that token amount is zero by comparing all singleTokenInfoParams
   let removal = [...cachedTokens];
   for (const singleTokenParam of singleTokenInfoParams) {
-    removal = removal.filter((token) => token.tokenAccountAddress !== singleTokenParam.tokenAccountAddress);
+    removal = removal.filter(
+      (token) =>
+        (token.metadata && !isValidWhitelistNft(singleTokenParam.identifiers, token.metadata)) ||
+        token.tokenAccountAddress !== singleTokenParam.tokenAccountAddress
+    );
```
haVincy, line 144: This is more like checking if any user already cached those non-whitelisted NFT.  Better do add the comment to describe why we add this check.

core/ui/src/public/Sell/index.tsx
```diff
@@ -129,7 +130,7 @@ export const Sell: React.FC<SellProps> = ({ wallet, walletConnectComponent, styl
 
   // fetch current wallet nfts when mount and when publicKey was changed.
   useEffect(() => {
-    if (!walletPublicKey) {
+    if (!walletPublicKey || shopLoading !== LoadStatus.Loaded) {
       return;
     }
```
haVincy, line 135: Like we discussed offline, the correct flow here is `progressiveLoadUserNFTs` should wait `fetchShopByShopAddress` to get the info.

core/sdk/src/vendor/token/fetchNftsFromWallet.ts
vanminh1701: okay, I will leave a comment for it.
This code is using for case: if your nfts are cached, then shop update allowSellAnyNft status to `not allow`, so not-whitelist nft will be added to remobal, then remove it from indexDB

core/ui/src/public/Sell/index.tsx
vanminh1701: yes, so what do you suggest? Current code is working right

haVincy: Let's add comment here as well like we discussed.

haVincy (approved):
Approved but make sure update the comments and tested.

PR #292, opened 2022-07-14
LIQ-689 update whitelist nft for auction

vanminh1701:


core/ui/src/public/Auction/CreateAuction.tsx
```diff
@@ -70,6 +70,15 @@ export const CreateAuction: React.FC<CreateAuctionProps> = ({
     [candyShop]
   );
 
+  const getShopIdentifiers = useCallback(async (): Promise<string[] | undefined> => {
+    if (shop?.allowSellAnyNft !== 0) return undefined;
```
RustySol, line 74: does it make sense to return `undefined` here? or should return `[]`?

RustySol: oh ok `fetchNftsFromWallet` takes `undefined` also

haVincy: @vanminhtran could you help to add comment above `getShopIdentifiers` here to explain `[]` vs `undefined` case. I think it will help us to maintain and also help users.

RustySol (approved):

haVincy (approved):

PR #300, opened 2022-07-22
[LIQ-697] Limit marketplace listings to 3 creators for spl shops

nibbus0x:


core/sdk/src/vendor/utils/validationUtils.ts
```diff
@@ -180,15 +186,22 @@ export const checkSettleParams = async (auction: web3.PublicKey, program: Progra
   }
 };
 
-export const checkCanExecSettle = async (
+export const checkCreators = async (
   treasuryMint: web3.PublicKey,
   nftMint: web3.PublicKey,
-  connection: web3.Connection
+  connection: web3.Connection,
+  isMarketplace: boolean
```
RustySol, line 193: suggest we can use ENUM here instead to indicate the transaction types

core/sdk/src/vendor/utils/validationUtils.ts
```diff
@@ -180,15 +186,22 @@ export const checkSettleParams = async (auction: web3.PublicKey, program: Progra
...
 ) => {
   const isNative = treasuryMintIsNative(treasuryMint);
   const [nftMetadata] = await getMetadataAccount(nftMint);
   const creators = await getNftCreators(nftMetadata, connection);
-  const creatorsLimit = isNative ? NATIVE_CREATORS_LIMIT : SPL_CREATORS_LIMIT;
+  const creatorsLimit = isMarketplace
+    ? isNative
```
RustySol, line 199: avoid usage of nested ternary condition

core/sdk/src/vendor/utils/validationUtils.ts
```diff
@@ -201,3 +213,15 @@ export const checkAHFeeAccountBalance = async (feeAccount: web3.PublicKey, conne
...
+const getCreatorLimit = (isNative: boolean, transactionType: TransactionType) => {
+  if (isNative && transactionType === TransactionType.Marketplace) {
+    return NATIVE_MARKETPLACE_CREATORS_LIMIT;
+  } else if (!isNative && transactionType === TransactionType.Marketplace) {
+    return SPL_MARKETPLACE_CREATORS_LIMIT;
+  } else if (isNative && transactionType == TransactionType.Auction) {
+    return NATIVE_AUCTION_CREATORS_LIMIT;
+  } else {
+    return SPL_AUCTION_CREATORS_LIMIT;
+  }
```
haVincy, line 226: 1. nit - `getCreatorsLimit`
2. Can do following?
```ts
switch (transactionType) {
   case TransactionType.Marketplace:
       return isNative ? NATIVE_MARKETPLACE_CREATORS_LIMIT : SPL_MARKETPLACE_CREATORS_LIMIT;
    case TransactionType.Auction:
        return isNative ? NATIVE_AUCTION_CREATORS_LIMIT : SPL_AUCTION_CREATORS_LIMIT
    default:
        // What's reasonable fall-back limit if not in above cases, seems should not happen.
        console.log(`Incorrect transationType ${transactionType} to get Creators Limit`);
}
    
```

RustySol: yes it should be `not reachable`

haVincy: Then should we throw Error for `default` case if it should not happen, otherwise, `getCreatorsLimit` will get nothing if we don't return something?

haVincy (approved):

PR #305, opened 2022-07-24
Impose limit on size of seller fee basis points a listed nft can have

nibbus0x:


haVincy (approved):
nit - Title seems missing a 't' for 'nft'

RustySol:
Will discuss more about the limits before we merge

PR #317, opened 2022-07-30
LIQ-727 auction activity UI

vanminh1701:


core/ui/src/public/AuctionActivity/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+  candyShop: CandyShop;
+  orderBy?: SortBy[] | SortBy;
+  statusFilters?: AuctionStatus[];
+  wallet?: AnchorWallet;
+}
+
+const LIMIT = 10;
+const DO_NOTHING_FUNC = () => {
+  // this prevent double call api transaction in useEffect and infinity lib
+};
```
haVincy, line 31: I'm seeing this empty function everywhere, i.e. `DO_NOTHING`. 
Should we extract them in `utils/array.ts` and rename `array.ts` to `helperFunc.ts`

RustySol:
should we base this PR on top of https://github.com/LIQNFT/candy-shop/pull/314 @vanminh1701 ?

PR #318, opened 2022-08-01
LIQ-729 Refresh UserNFT once creadit card payment succeed

vanminh1701:


core/ui/src/constant/index.ts
```diff
@@ -7,7 +7,7 @@ export enum LoadStatus {
 }
 export const TIMEOUT_EXTRA_LOADING = 3_000;
 export const POLLING_SHOP_INTERVAL = 3_000;
-export const POLLING_INTERVAL = POLLING_SHOP_INTERVAL + 500;
+export const POLLING_INTERVAL = 1000;
```
haVincy, line 10: Why need to change this?

vanminh1701: This is timeout for client to check latest localStorage. I saw we have a `delay time` after API signal reponsed and updated localStorage. If we descrease `POLLING_INTERVAL`, mean we will descrease `delay time`
I don't remember why I didn't intermediatelly refresh component at that time. Refactor it for now is not usefull. This flow will be refactor when I integrate WebSocket to project

haVincy: kk

haVincy (approved):

PR #321, opened 2022-08-02
LIQ-728 add filter order by collection id

vanminh1701:


RustySol (approved):

core/ui/src/public/Orders/index.tsx
```diff
@@ -46,6 +46,10 @@ interface OrdersProps {
 
 /**
  * React component that displays a list of orders
+ * @param filters:
+ *    - true: list collections from that current shop
+ *    - CollectionFilter: hardcode collections
+ *    - auto: list collection from Shop filter UI, prop shopFilters=true
```
haVincy, line 52: It seems we don't really need 'auto' here, if  `filters` is `undefined`, we can treat it as auto?

haVincy: As offline synced, @vanminhtran let's add TODO comments to indicate we will remove the hardcode filter and enhance this to better usages.

PR #323, opened 2022-08-02
[core/types] add drop schema

dongnguyenltqb:


core/types/src/response/drop.ts
```diff
@@ -0,0 +1,37 @@
+import { NftAttribute, NftProperties } from '../nft';
+
+export enum VaultStatus {
+  CREATED = 0,
+  SALE_STARTED = 1,
+  WHITELIST_STARTED = 2,
```
RustySol, line 6: lets swap the order of `WHITELIST_STARTED` and `SALE_STARTED`

core/types/src/response/drop.ts
```diff
@@ -0,0 +1,37 @@
...
+  SALE_COMPLETED = 3
+}
+
+export interface Drop {
+  vaultAddress: string;
+  type: number;
```
RustySol, line 12: make this a enum

```
enum DropType {
   Editioned = 0
}
```

core/types/src/response/drop.ts
```diff
@@ -0,0 +1,37 @@
...
+  shopId: string;
+  price: string;
+  maxSupply: number;
+  currentSupply: number;
+  startTime: string;
+  whitelistTime: string;
```
RustySol, line 18: can be null

core/types/src/response/drop.ts
```diff
@@ -0,0 +1,37 @@
...
+  whitelistMint: string;
```
RustySol, line 19: can be null

RustySol (approved):

RustySol:
will merge with the bump commit

PR #326, opened 2022-08-04
Fix isEnterprise flag checking

RustySol:


core/cli/src/cli.ts
```diff
@@ -50,7 +50,7 @@ programCommand('sellMany')
       settings: {
         mainnetConnectionUrl: rpcUrl
       },
-      isEnterprise: isEnterprise ? true : false
+      isEnterprise: isEnterprise === 'true' ? true : false
```
haVincy, line 53: It seems can make it a general method and use it everywhere.
```ts
const isEnterprise = (isEnterpriseArg: string): boolean => {
  return isEnterprise === 'true';
}
```

```ts
let { ..., isEnterpriseArg } = cmd.opts();

const candyShop = new CandyShop({
    ...
    isEnterprise: isEnterprise(isEnterpriseArg);
})
...
```

PR #333, opened 2022-08-11
Feat/edition drop e2e

RustySol:


core/sdk/package.json
```diff
@@ -10,6 +10,7 @@
   },
   "dependencies": {
     "@liqnft/candy-shop-types": "0.2.23",
+    "@metaplex-foundation/js": "^0.14.0",
```
RustySol, line 13: @nibbus0x  possible we move this to `devDependencies`? Cuz this is only used for test cases

RustySol: will fix in another PR.

PR #334, opened 2022-08-11
Core/edition drop

vanminh1701:


vanminh1701:
test link - https://core-edition-drop.d22tqykss69onq.amplifyapp.com/

core/ui/src/utils/helperFunc.ts
```diff
@@ -26,3 +26,14 @@ export function removeDuplicate<T>(oldList: T[], addList: T[], key: keyof T): T[
 export function EMPTY_FUNCTION(): void {
   //
 }
+
+export function calculateRemainTimeValue(countdown: number): string {
```
haVincy, line 30: I like this extraction. Would it be better to call it `getCountdownTimerString` ?

core/ui/src/hooks/useObserver.ts
```diff
@@ -0,0 +1,49 @@
...
+      });
+    };
+
+    const options = {
+      root: null,
+      rootMargin: '0px',
+      threshold: 1.0
+    };
+
+    const observer = new IntersectionObserver(callbackObserver, options);
```
haVincy, line 42: Not a comment.

Better to check if current browser supports Intersection Observer or not.  
Not all modern & useful webAPIs can be used in JS on all browsers/platforms without condition.

```ts
if ('IntersectionObserver' in window)
```
This is not so important for our level as we don't expect to support many devices yet, just a suggestion that might help you to avoid pitfalls or unexpected errors in future dev when on bigger product.

core/ui/src/hooks/useObserver.ts
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    let prevRatio = 0;
+    const callbackObserver = (entries: any[]) => {
+      entries.forEach((entry) => {
+        console.log('entry.intersectionRatio', entry.intersectionRatio);
+        if (entry.intersectionRatio > prevRatio) {
+          callbackFnRef.current && callbackFnRef.current();
+          window.scrollTo({ top: window.scrollY - 50, behavior: 'smooth' });
+        }
```
haVincy, line 31: Suggest to make '50' a constant in this hook.

core/ui/src/utils/format.ts
```diff
@@ -4,12 +4,12 @@ dayjs.extend(utc);
...
-  auctionMinute: string;
+type GetTimeType = {
+  hour: string;
+  minute: string;
   clockFormat: string;
-  startNow?: boolean;
-  startDate: string;
+  isNow?: boolean;
+  date: string;
 };
```
haVincy, line 13: 1. Can use `interface` here and just call it `TimerInfo`
2. The file name seems can rename to `Timer.ts` and move `shortenAddress` method to `helperFunc.ts`.

core/ui/src/utils/getPrice.ts
```diff
@@ -1,15 +1,14 @@
...
 
 export const getPrice = (
   shopPriceDecimalsMin: number,
   shopPriceDecimals: number,
-  order?: OrderSchema,
+  price: string | number | undefined,
   exchangeInfo?: ShopExchangeInfo
 ): any => {
-  if (!order?.price || !exchangeInfo) return null;
+  if (!price || !exchangeInfo) return null;
```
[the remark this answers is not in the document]
vanminh1701: We had logic to check null value from getPrice

core/sdk/src/vendor/token/fetchNftsFromWallet.ts
```diff
@@ -211,3 +213,78 @@ const getValidChunkSize = (chunkSize: number | undefined): number => {
...
+      account: token.account,
+      uri: undefined,
+      tokenPubkey: token.pubkey.toString(),
+      tokenMint: token.account.data.parsed.info.mint,
+      amount: token.account.data.parsed.info.tokenAmount.amount,
+      metadata: undefined,
+      edition: undefined
+    };
+  });
+  // console.log('tokens', tokens);
```
RustySol, line 243: remove this

core/ui/src/utils/ErrorHandler.ts
```diff
@@ -3,10 +3,13 @@ import { CandyShopError } from '@liqnft/candy-shop-sdk';
 
 export enum ErrorType {
   InvalidWallet = 'InvalidWallet',
-  GetAccountInfoFailed = 'GetAccountInfoFailed'
+  GetAccountInfoFailed = 'GetAccountInfoFailed',
+  RequestExceedLimit = 'RequestExceedLimit'
```
RustySol, line 7: this error doesnt seem used?

core/sdk/src/vendor/token/fetchNftsFromWallet.ts
```diff
@@ -211,3 +213,78 @@ const getValidChunkSize = (chunkSize: number | undefined): number => {
   }
   return chunkSize;
 };
+
+export const fetchUserMasterNFTs = async (
+  userWalletPublicKey: web3.PublicKey,
+  connection: web3.Connection
+): Promise<any[]> => {
```
haVincy, line 220: Should we make an `interface` for it?

core/sdk/src/vendor/token/fetchNftsFromWallet.ts
```diff
@@ -211,3 +213,78 @@ const getValidChunkSize = (chunkSize: number | undefined): number => {
...
+  const [rawTokens, nftsInfo] = await Promise.all([
+    connection.getParsedTokenAccountsByOwner(userWalletPublicKey, {
+      programId: TOKEN_PROGRAM_ID
+    }),
+    safeAwait(fetchNftsFromWallet(connection, userWalletPublicKey, undefined))
+  ]);
+
+  if (!rawTokens.value || !nftsInfo.result) return [];
```
haVincy, line 228: Seems we want to return empty array If any error occurs, can do `safeAwait` on `Promise.all` directly. 

```ts
const tokens = await safeAwait(Promise.all[
   ...
]);
if (tokens.error || !tokens.result) {
  // Debug error log
  console.log()
  return [];
}

const [rawTokens, nftsInfo] = tokens.result;
  
```

core/sdk/src/vendor/token/fetchNftsFromWallet.ts
```diff
@@ -211,3 +213,78 @@ const getValidChunkSize = (chunkSize: number | undefined): number => {
...
+  // console.log('tokens', tokens);
+  const tokenInfoMap: Map<string, TokenInfo> = new Map();
+  for (const token of tokens) {
+    if (Number(token.amount) > 0) {
+      tokenInfoMap.set(token.tokenPubkey, token);
+    }
+  }
+
+  const nftsInfoMap = new Map<string, SingleTokenInfo>();
+  if (!nftsInfo.result) return []; // handle error
```
haVincy, line 252: This should not happen if we already handle the error case at line 228?

core/sdk/src/vendor/token/fetchNftsFromWallet.ts
```diff
@@ -211,3 +213,78 @@ const getValidChunkSize = (chunkSize: number | undefined): number => {
...
+      if (newEditionAccountInfo) {
+        const newEditionInfo = parseMasterEditionV2((newEditionAccountInfo as any).data);
+        const editionInfo = parseEdition((newEditionAccountInfo as any).data);
+        return {
+          ...e,
+          maxSupply: newEditionInfo.maxSupply?.toString(),
+          edition: editionInfo.edition.toString(),
+          supply: newEditionInfo.supply.toString()
+        };
+      }
```
haVincy, line 280: We can do nothing if the `newEditionAccountInfo` is `undefined`?

core/sdk/src/vendor/token/parseData.ts
```diff
@@ -198,3 +198,25 @@ export const parseEditionMarker = (buffer: Buffer): EditionMarker => {
 export const parseEdition = (buffer: Buffer) => {
   return deserializeUnchecked(METADATA_SCHEMA, Edition, buffer) as Edition;
 };
+
+export interface TokenInfo {
```
haVincy, line 202: Is it for the EditionDrop? Can we name it more specific, not just only `TokenInfo`, might make confusion between `SingleTokenInfo` (although I know this is a bad naming as well but it already shipped, can not do anything.)

core/sdk/src/CandyShopDrop.ts
```diff
@@ -102,7 +102,6 @@ export abstract class CandyShopDrop {
     } = params;
 
     const [vaultAccount] = await getEditionVaultAccount(candyShop, nftOwnerTokenAccount);
-
```
haVincy, line 105: Remove this ineffective difference.

core/sdk/src/vendor/token/parseData.ts
vanminh1701: I will name it to `RawTokenInfo` for the same meaning with function fetching it

RustySol (approved):

haVincy (approved):

PR #361, opened 2022-08-22
LIQ-772: update format query nft attribute

vanminh1701:


core/types/src/nft/nft.ts
```diff
@@ -15,6 +15,5 @@ export interface NftProperties {
   creators: NftCreator[];
 }
 export interface NftAttribute {
-  value: string;
-  trait_type: string;
+  [trait_type: string]: string;
```
RustySol, line 18: is this used in the backend API response? @dongnguyenltqb

RustySol (approved):

PR #373, opened 2022-08-25
types/expose order nft attributes

RustySol:


RustySol:
Will merge after the `candy-shop` release

PR #375, opened 2022-08-26
Poc/websocket

vanminh1701:


example/constant/formConfiguration.ts
```diff
@@ -12,6 +12,7 @@ export const DEFAULT_FORM_CONFIG = {
     mainnetConnectionUrl: 'https://ssc-dao.genesysgo.net/'
   }),
   paymentProvider: JSON.stringify({
-    stripePublicKey: ''
+    stripePublicKey:
+      'pk_test_51L3BeuJr9YyWR0YtVYfi3NK7U9iKavPRtUNz1kRkvZjjOsIC2qoZd4JOOnO9TDU22mKg2UMmgVKyDYscP2o7sv0800A1tCggV5'
```
haVincy, line 16: Should remove this.

core/ui/src/constant/SocketEvent.ts
```diff
@@ -0,0 +1,31 @@
...
+  try {
+    return JSON.parse(message);
+  } catch (error) {
+    console.log('Parse socket message error=', error);
+  }
+}
+
+export function stringifyMessage(data: { event: EventName; data: any }): string {
+  return JSON.stringify(data);
+}
```
haVincy, line 31: I feel we can have a folder for WebSocket, `core/ui/src/socket` to gather those WebSocket related files/codes.

What is `JoinLeaveRoomType` here means? The wording seems like the common websocket usages for messenger app. We can simply call it `sessionId` & `StartAbortSessionType`.

vanminh1701: okay, `room` is a concept from socket.io, and you can join many rooms at the same time. So just keep it, I will rename `JoinLeaveRoomType`

core/ui/src/constant/SocketEvent.ts
```diff
@@ -0,0 +1,32 @@
...
+  auctionUpdated = 'auctionUpdated',
+  auctionUpdateBid = 'auctionUpdateBid',
+  auctionUpdateStatus = 'auctionUpdateStatus',
+  getAuctionByAddressAndWallet = 'getAuctionByAddressAndWallet',
+  dropCreatedOrUpdated = 'dropCreatedOrUpdated'
+}
+
+export type JoinLeaveRoomType = {
+  subscribeId: string;
+};
```
haVincy, line 20: `StartStopSessionType` & `sessionId` ?

core/ui/src/public/Context/Socket.tsx
```diff
@@ -0,0 +1,118 @@
+import React, { createContext, ReactElement, useCallback, useContext, useEffect, useState } from 'react';
+import { parseSocketMessage, EventName, stringifyMessage } from 'constant/SocketEvent';
+
+const createSocket = (): WebSocket => new WebSocket('ws://localhost:8084');
+
+interface SocketContextData {
+  socket?: WebSocket;
+  onListenEvent: (event: EventName, callback: (data: any) => void) => AbortController;
```
haVincy, line 8: How about `onServerEventEmits` ? It's more descriptive for client WebSocket usages.

core/ui/src/public/Orders/index.tsx
```diff
@@ -191,15 +175,39 @@ export const Orders: React.FC<OrdersProps> = ({
...
+  }, [fetchOrders]);
+
+  useEffect(() => {
+    const controllers = [
+      onListenEvent(EventName.orderOpened, (order: Order) => {
+        setOrders((list) => {
+          const newList = removeDuplicate<Order>([order], list, 'tokenMint');
+
+          const { column, order: sortOrder } = sortedByOption.value as { column: keyof Order; order: 'asc' | 'desc' };
+          const _sortFunc = (a: Order, b: Order) =>
```
haVincy, line 188: Remove `_` prefix, it's not recommended naming in TS, unless it's private member in Class and we have getter method for it.

- https://google.github.io/styleguide/tsguide.html#identifiers

core/ui/src/components/Drop/List/index.tsx
```diff
@@ -104,6 +109,22 @@ export const Drops: React.FC<DropsProps> = ({ candyShop, wallet, walletConnectCo
       });
   };
 
+  // socket
+  useEffect(() => {
+    const controllers = [
```
haVincy, line 114: Curious, we declare the `controllers` here, wouldn't the reference will be changed every time once component renders, so the purpose here is we need to add the listeners and clean the listeners every time when component is being mounting / unmounting?

vanminh1701: actually, removeControllers will be run every time useEffect re-run. If we dont remove controllers at that time, the listeners are stil there and we will have more than 1 listeners for 1 event, the result maybe we will add duplicate item to list and display on UI, or some unexpected results. So better just clean it every useEffect re-run or unmount phase.

core/ui/src/public/Context/Socket.tsx
haVincy: okay

haVincy:
Test link - https://poc-websocket.d22tqykss69onq.amplifyapp.com/

haVincy (approved):

core/ui/src/components/Drop/List/index.tsx
haVincy: okay, that sounds good and it's the desired result we want to avoid heap grows.

PR #388, opened 2022-09-07
[LIQ-809] Make Edition update authority same as Master

RustySol:


core/sdk/src/factory/program/v2/editionDrop/enterpriseMintPrint.ts
```diff
@@ -111,6 +115,14 @@ export const mintPrint = async (
     getEditionMarkAccount(masterMint, editionNumber.toNumber())
   ]);
 
+  const metadataAccount = await safeAwait(program.provider.connection.getAccountInfo(masterEditionMetadata));
+  if (metadataAccount.error || !metadataAccount.result) {
+    throw new CandyShopError(CandyShopErrorType.NotReachable);
+  }
+
+  const metadata = parseMetadata(metadataAccount.result!.data);
+  const updateAuthority = new PublicKey(metadata.updateAuthority);
```
nibbus0x, line 124: Maybe make a function that gets `updateAuthority` so that it can be used in both files.

PR #394, opened 2022-09-13
Update default shop config

RustySol:


RustySol:
Upgrade the default shop version to v2

PR #412, opened 2022-10-08
[sdk] update ATA programId to redeemDrop

RustySol:


RustySol:
There are situations where after users commit a master edition NFT, they will close the token account. A change is added in the onchain program to make sure the check will be conducted against the mint account address directly instead of relying on the mint address in the token account. This change adds an extra ATA program argument into the `redeemNft` rpc handler, so the committer's master edition NFT token account can be re-initialized upon redemption.

RustySol:
close for https://github.com/LIQNFT/candy-shop/pull/414

PR #417, opened 2022-10-11
fix: update OrderDetail union types

neversaynever0502:
Currently, the main UI components: OrderDetail in public all need candyShop: CandyShop as prop but the methods they called from sdk now can use without the CandyShop instance but just currencySymbol or partial infos. 

To avoid breaking backward compatibility, I use union-types that can serve OrderDetail with partial shop info without CandyShop instance.


core/ui/src/public/OrderDetail/index.tsx
```diff
@@ -70,8 +88,7 @@ export const OrderDetail: React.FC<OrderDetailProps> = ({
 
     if (order && !nftInfo) {
       setLoadingNftInfo(true);
-      candyShop
-        .nftInfo(order.tokenMint)
+      fetchNFTByMintAddress(order.tokenMint)
```
haVincy, line 91: Nice catch.

haVincy (approved):
Verify if no issue before merge.

RustySol (approved):
LGTM, will merge once pass QA

PR #423, opened 2022-10-14
[sdk] extract the usages for Auction

RustySol:


RustySol:
Pending on refactoring code inside `CandyShop.ts` to use the methods from `CandyShopAuction.ts`

neversaynever0502 (approved):

PR #433, opened 2022-10-28
[ui] fix-eth-bugs

RustySol:


RustySol:
- Update Activity SOL link
- Fix config setting
- Fix Stat number
- Update Eth ExplorerLink
- Update Eth Activity

neversaynever0502:
We used explorerLink props in component ExplorerLink.
In blockchain SOL and Activity Component,  we need explorerLink props in three solscan icons to recognize which one connect to each one. 

But in blockchain Eth, we can just use env to determine the link url in component ExplorerLink.

core/ui/src/public/Activity/index.tsx
```diff
@@ -118,30 +137,56 @@ export const Activity: React.FC<ActivityProps> = ({ identifiers, orderBy, candyS
...
+                          <ExplorerLink
+                            type="tx"
+                            address={trade.txHashAtCreation}
+                            candyShopEnv={candyShop.env}
+                            explorerLink={ExplorerLinkBase.Explorer}
+                          >
+                            <IconExplorer />
+                          </ExplorerLink>
+                        </>
+                      ) : candyShop.env == Blockchain.Eth || candyShop.env == Blockchain.EthTestnet ? (
```
RustySol, line 167: suggest we use this function `etherumEnvIconTrans` it will reduce code duplicates as we support more EVM chains

core/ui/src/public/Activity/index.tsx
```diff
@@ -85,6 +89,19 @@ export const Activity: React.FC<ActivityProps> = ({ identifiers, orderBy, candyS
     return () => controller.abort();
   }, [onSocketEvent]);
 
+  const etherumEnvIconTrans = (env: Blockchain) => {
+    switch (env) {
+      case Blockchain.Eth:
+        return <IconPolygon />;
```
RustySol, line 95: IconEth?

core/ui/src/public/Activity/index.tsx
neversaynever0502: OK, I have used etherumEnvIconTrans function to replace it.

core/ui/src/public/Activity/index.tsx
neversaynever0502: Yes, fixed.

vanminh1701 (approved):

PR #434, opened 2022-10-29
[core] Refactor ui/src/services and remove uncommon getters in BaseShop

haVincy:


haVincy:
Test link - https://core-refactor-multi-chain-codes.d22tqykss69onq.amplifyapp.com/
This PR is for Phase 1 refactoring, will have Phase 2 refactoring that mainly in core/sdk.

core/ui/src/public/Auction/Auctions.tsx
```diff
@@ -23,14 +23,10 @@ interface AuctionsProps extends ShopProps {
...
   statusFilters = DEFAULT_LIST_AUCTION_STATUS,
-  blockchain,
   candyShop,
   wallet
 }) => {
-  const auctioneer = useMemo(
-    () => AuctioneerFactory({ candyShop, blockchain, wallet }),
-    [candyShop, blockchain, wallet]
-  );
+  const store = useMemo(() => StoreProvider({ candyShop, wallet }), [candyShop, wallet]);
```
haVincy, line 29: @beepsaidblue @vanminh1701  I was wondering if it's good that we add check here to avoid `EthCandyShop` to use `Auctions`. What do you guys think?
```ts
if (!(candyShop instanceof CandyShop)) {
    // return something empty component
}
```

haVincy: NVM, I just realize if the shopAddress doesn't have Auction data, it's will be empty. I think maybe no need to add this redundant codes.

vanminh1701 (approved):

RustySol (approved):

PR #440, opened 2022-11-01
[sdk] Refactor the API request that with query string

haVincy:


core/sdk/src/factory/backend/utils.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Get the parameterize string of single query / multiple queries
+ * @param queryString string or array of string
+ * @returns string with parameter for sending request to server
+ */
+export const getParametrizeQuery = (queryString: string | string[]): string => {
```
vanminh1701, line 6: curious, do we have `multiple queries` in our repo ?

haVincy: Not yet, just write it as extendability in case needed.

core/sdk/src/factory/backend/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+  if (singleQuery) {
+    return `?${queryString}`;
+  }
+  // Multiple queries case
+  const firstQuery = queryString[0];
+  let followQueries = '';
+  for (let i = 1; i < queryString.length; i++) {
+    followQueries = followQueries.concat(`&${queryString[i]}`);
+  }
+  return `${firstQuery}${followQueries}`;
```
haVincy, line 20: a bug, `firstQuery` should have `?` prefix.

vanminh1701 (approved):

PR #446, opened 2022-11-04
[ui] get current NFTs from trade websocket

neversaynever0502:
LIQ-918


haVincy (approved):
LGTM, but need @vanminh1701 to double check then we can merge.

vanminh1701 (approved):

PR #447, opened 2022-11-04
[core] add linter

neversaynever0502:
LIQ-917


package.json
```diff
@@ -48,5 +49,13 @@
...
     "react-dom": "./core/ui/node_modules/react-dom/profiling",
     "scheduler/tracing": "./core/ui/node_modules/scheduler/tracing-profiling"
+  },
+  "dependencies": {
+    "@typescript-eslint/eslint-plugin": "^4.14.0",
+    "@typescript-eslint/parser": "^4.14.0",
+    "eslint": "^7.18.0",
+    "eslint-config-prettier": "^8.5.0",
+    "eslint-plugin-prettier": "^4.0.0",
+    "eslint-plugin-react-hooks": "^4.4.0"
```
haVincy, line 59: Should use `devDependencies` ?

package.json
```diff
@@ -19,7 +19,8 @@
     "clean:sdk": "rm -rf core/sdk/node_modules && rm -rf core/sdk/dist",
     "clean:types": "rm -rf core/types/node_modules && rm -rf core/types/dist",
     "clean:ui": "rm -rf core/ui/node_modules && rm -rf core/ui/dist",
-    "clean:all": "yarn clean:root && yarn clean:sdk && yarn clean:ui && yarn clean:types"
+    "clean:all": "yarn clean:root && yarn clean:sdk && yarn clean:ui && yarn clean:types",
+    "lint": "eslint \"core/{sdk,cli,types,ui}/**/*.ts\""
```
haVincy, line 23: If we are adding this in yarn script, do we need to enable pre-push hook and CI to check it? @RustySol

haVincy:
If we gonna make this linter, I think we better do the lint fix on all codes in the same PR as coverage.

package.json
RustySol: Yes that will be the plan. @neversaynever0502  pushed this up for our early review.

package.json
neversaynever0502: Yes, I have changed it.

example/eth/ShopConfig.tsx
```diff
@@ -83,7 +83,7 @@ export const ShopConfig: React.FC<ShopConfigProps> = ({ setCandyForm, candyForm
 
   useEffect(() => {
     if (candyForm) onUpdateFormState(candyForm);
-  }, [candyForm]);
+  }, [candyForm, onUpdateFormState]);
```
RustySol, line 86: this change seems not related to linting

RustySol (approved):

example/eth/ShopConfig.tsx
RustySol: turns out it's related to `react-hooks/exhaustive-deps`

vanminh1701 (approved):

vanminh1701:
nice config for projects 👍

PR #462, opened 2022-11-09
[sdk] Refactor ApiCaller in EthereumSDK

vanminh1701:


core/sdk/src/factory/conveyor/eth/conveyor.ts
```diff
@@ -14,17 +14,9 @@ import {
   SplitReceiver
 } from './types';
 
-const Logger = 'EthereumSDK';
-
 export class EthereumSDK {
-  private readonly apiPath = '/eth';
```
haVincy, line 20: I feel we can still keep this path as constant for axios to use?

RustySol (approved):

PR #476, opened 2022-11-19
[sdk] Refactor configBaseUrl flow in CandyShop & EthCandyShop

haVincy:


core/sdk/src/shop/base/BaseShop.ts
```diff
@@ -98,8 +96,11 @@ export abstract class BaseShop {
     this._programId = params.programId;
     this._settings = params.settings;
     this._baseUnitsPerCurrency = Math.pow(10, this._settings.currencyDecimals);
-    this.baseUrl = getBaseUrl(this.env);
```
haVincy, line 101: Remove this protected member since don't need it anymore in inheritors after cleaned up `ApiCaller` for `EthCandyShop` .

core/sdk/src/shop/base/BaseShop.ts
```diff
@@ -98,8 +96,11 @@ export abstract class BaseShop {
...
-    configBaseUrl(this.baseUrl);
+  }
+
+  // Only inheritors can call this configuration
+  protected static configEndpoint(env: Blockchain) {
+    configBaseUrl(getBaseUrl(env));
```
haVincy, line 103: I can not think of better way without doing massive refactoring on all current API methods, so just refactored with simplest way to gathering the usage in `BaseShop` for `CandyShop` and `EthCandyShop`.

RustySol (approved):

PR #478, opened 2022-11-20
[fix] error api response

vanminh1701:


RustySol (approved):

haVincy (approved):
I feel this is needed to be highlighted in release notes in builders channel for those who is using SDK. @RustySol

PR #479, opened 2022-11-22
[feat] add css variables

vanminh1701:


haVincy:
also include @neversaynever0502 as reviewer to be aware of this.

core/ui/src/variable.less
```diff
@@ -1,21 +1,37 @@
-@primary-color: #7522f5;
-@popover-background: #1a2029;
-@primary-color-light: #f4edfe;
-@primary-color-dark: #4c00c0;
-@secondary-color-dark: #e9ae00;
+//Color
+@candy-color-primary: var(--candy-color-primary, #7522f5);
```
haVincy, line 2: How about we 
1. Clearly name the `cany-shop-xxx-xxx`
2. Use  `liqnft-cs-xxx-xxx` , it's uncommon abbreviation but reasonable to ourselves.

vanminh1701: can do and I think we should merge some vars to have less configs and I think less color in our site will better for UX

haVincy: Agree, any key example you can illustrate?

haVincy:
nit - commit title should use [ui]

core/ui/src/variable.less
RustySol: lets use `cany-shop-xxx-xxx`

haVincy: Updated new commit with `candy-shop-xxx`,

haVincy (approved):

RustySol (approved):

PR #482, opened 2022-11-26
[fix] order detail not found

vanminh1701:
[fix] remove Stripe error

[fix] auction modal


core/ui/src/components/BuyModal/index.tsx
```diff
@@ -103,92 +106,93 @@ export const BuyModal: React.FC<BuyModalProps> = ({
...
-        throw error;
-      });
-    }
-    const fiatPriceRes = await CandyShopPay.getTokenFiatMoneyPrice(
-      { shopId: shopAddress, tokenAccount: order.tokenAccount },
-      { quoteCurrencyType: PaymentCurrencyType.USD }
-    ).catch((error: Error) => {
-      throw error;
-    });
+  /** Comment out Stripe payment feature */
```
haVincy, line 109: Why we need to comment out Stripe payment here? 
We can check If doesn't have Stripe publicKey, all stripe codes should not be used.

RustySol:
Looks like this PR should just focus on fixing the auction modal.

RustySol (approved):

PR #488, opened 2022-12-05
[types][sdk] merge drop and order api

dongnguyenltqb:
[LIQ-988]

[LIQ-988]: https://liqnft.atlassian.net/browse/LIQ-988?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/types/src/response/order.ts
+++ b/core/types/src/response/order.ts
@@ -1,4 +1,5 @@
 import { NftAttribute, NftCreator, NftProperties } from '../nft';
+import { Drop } from './drop';
 
 export enum Side {
   BUY,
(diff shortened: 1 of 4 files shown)


core/types/src/response/order.ts
```diff
@@ -57,3 +58,7 @@ export interface Order {
   symbol?: string;
   decimals: number;
 }
+
+export type OrderOrDrop = (Order | Drop) & {
+  entity: 0 | 1;
+};
```
haVincy, line 64: Why we need this `entity` ?

PR #489, opened 2022-12-05
[example] fix order not found notification

vanminh1701:


core/ui/src/public/OrderDetail/index.tsx
```diff
@@ -45,40 +45,39 @@ export const OrderDetail: React.FC<OrderDetailProps> = ({
...
   const publicKey = wallet?.publicKey?.toString();
   const isUserListing = publicKey && order && order.walletAddress === publicKey;
+  const candyShopAddress = candyShop.candyShopAddress;
 
   useEffect(() => {
-    if (!order) {
-      setLoadingOrder(true);
+    if (!tokenMint || !candyShopAddress) return;
+    setLoadingOrder(true);
+    fetchOrderByShopAndMintAddress(candyShopAddress, tokenMint)
```
haVincy, line 53: Doesn't the `SolStore.ts#getOrderNft` is equivalent to `fetchOrderByShopAndMintAddress` ? It is using `CandyShop#activeOrderByMintAddress`.

Are we fixing this for EthStore case? If yes, the feasible fix should update the `EthStore.ts`, let it use `fetchOrderByShopAndMintAddress`.

vanminh1701: the root cause is not from Store, but react-router-dom (https://v5.reactrouter.com/web/api/Route/route-render-methods. I updated PR.

core/ui/src/public/Activity/index.tsx
```diff
@@ -73,6 +73,9 @@ export const Activity: React.FC<ActivityProps> = ({ identifiers, orderBy, candyS
   );
 
   useEffect(() => {
+    setOffset(0);
+    setHasMore(true);
+    setTrades([]);
     getTrades(0, LIMIT)();
```
vanminh1701, line 79: this help to reset UI when candyShop cofig is changed

haVincy (approved):

PR #492, opened 2022-12-11
[sdk] Refactor connectionUrl for ShopSettings

haVincy:
Make the connectionUrl can not be undefined, it should be provided
in shop init stage.


core/sdk/src/shop/sol/CandyShop.ts
```diff
@@ -212,7 +206,7 @@ export class CandyShop extends BaseShop implements CandyShopAuctioneer, CandySho
     }
 
     this._candyShopAddress = candyShopAddress;
-    this._isEnterprise = isEnterprise ? true : false;
+    this._isEnterprise = !!isEnterprise; // Not Not operators to parse undefined as false
```
vanminh1701, line 209: suggest to use `Boolean()`, then no need this comment

vanminh1701 (approved):

core/sdk/src/shop/sol/CandyShop.ts
haVincy: Okay, let me update

PR #501, opened 2022-12-15
[sdk] mint print with info

RustySol:


RustySol:
close for https://github.com/LIQNFT/candy-shop/pull/533

PR #504, opened 2022-12-20
[ui] add new variable

neversaynever0502:
How to test:
1. "yarn start" to run the example
2. go into core/ui and "yarn start" to listen the ui realtime changes
3. change theme variable in core/ui/src/variable.less
4. the example ui will be changed in the meantime


core/ui/src/assets/IconChevronUp.tsx
```diff
@@ -1,11 +1,12 @@
 import React from 'react';
+// import '../variable.less'
```
RustySol, line 2: let's remove this?

RustySol (approved):

core/ui/src/assets/IconChevronUp.tsx
neversaynever0502: Yes, done.

core/ui/src/components/PoweredBy/style.less
```diff
@@ -30,4 +32,6 @@
   svg {
     margin-left: 8px;
   }
+
+  --IconCandyShopOutlined: @candy-shop-color-primary;
```
haVincy, line 36: CSS variables tend to use lower cases.

Seems we will use the same color `@candy-shop-color-primary` for `IconCandyShop` & `IconCandyShopOutlined` & Chevron at this moment. 

We can just declare one css variable in root? 

```less
.root {
   --candy-shop-activate-icon-color: @candy-shop-color-primary
}
```

In various places for SVG:
```tsx
fill="var(--candy-shop-activate-icon-color)"
```

neversaynever0502:
```
:root {
   --candy-shop-activate-icon-color: @candy-shop-color-primary
}
```
added, worked fine.
also changed other css variable to use lower cases.

vanminh1701 (approved):

RustySol (approved):

PR #518, opened 2023-01-09
[sdk] add API for fulfillOrder payload

RustySol:
LIQ-1191

### DIFF
--- a/core/sdk/src/factory/conveyor/eth/EthereumPort.ts
+++ b/core/sdk/src/factory/conveyor/eth/EthereumPort.ts
@@ -9,7 +9,8 @@ import {
   SeaportOrderData,
   ShopResponse,
   SplitReceiver,
-  ExecuteOrderParams
+  ExecuteOrderParams,
+  OrderPayloadResponse
 } from './types';
 
 interface CreateOrderParams {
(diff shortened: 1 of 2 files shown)


core/sdk/src/factory/conveyor/eth/types/order.type.ts
```diff
@@ -37,8 +37,15 @@ export interface SeaportOrderData {
 
 export interface OrderResponse {
   uuid: string;
+  shopUuid: string;
   rawOrderParams: string;
   signature: string;
+  createdAt: Date;
```
vanminh1701, line 43: This field type should be `string`, because we don't have any middleware to convert response `string` to `Date`

core/sdk/src/factory/conveyor/eth/EthereumPort.ts
```diff
@@ -133,6 +134,12 @@ export class EthereumPort {
     return axiosInstance.get(`${this.apiPath}/order/${uuid}`, {}).then((res) => res.data.result);
   }
 
+  async getFulfillOrderPayloadByUuid(uuid: string, buyerAddress: string): Promise<OrderPayloadResponse> {
+    return axiosInstance
+      .get(`${this.apiPath}/order/payload/${uuid}/buyer/${buyerAddress}`, {})
```
vanminh1701, line 139: seem no need 2nd argument: `{}`

vanminh1701 (approved):

PR #519, opened 2023-01-09
[sdk] add getNftPurchasePayload

RustySol:
LIQ-1191

### DIFF
--- a/core/sdk/src/factory/conveyor/eth/EthereumPort.ts
+++ b/core/sdk/src/factory/conveyor/eth/EthereumPort.ts
@@ -126,17 +126,17 @@ export class EthereumPort {
     };
   }
 
-  async getShopByUuid(uuid: string): Promise<{ result: ShopResponse }> {
-    return axiosInstance.get(`${this.apiPath}/shop/${uuid}`).then((res) => res.data);
+  getShopByUuid(shopUuid: string): Promise<{ result: ShopResponse }> {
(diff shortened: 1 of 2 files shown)


core/sdk/src/factory/conveyor/eth/EthereumPort.ts
```diff
@@ -126,17 +126,17 @@ export class EthereumPort {
     };
   }
 
-  async getShopByUuid(uuid: string): Promise<{ result: ShopResponse }> {
+  getShopByUuid(uuid: string): Promise<{ result: ShopResponse }> {
```
haVincy, line 129: Does this needs `shopUuid` instead of `uuid` in OrderPayload?
If so, should we rename the param as `shopUuid`.

core/sdk/src/factory/conveyor/eth/EthereumPort.ts
```diff
@@ -126,17 +126,17 @@ export class EthereumPort {
...
     return axiosInstance.get(`${this.apiPath}/shop/${uuid}`).then((res) => res.data);
   }
 
-  async getOrderByUuid(uuid: string): Promise<OrderResponse> {
+  getOrderByUuid(uuid: string): Promise<OrderResponse> {
```
haVincy, line 133: Similar question here, does the `uuid` here requires `shopUuid` or `uuid` ?

haVincy (approved):

PR #520, opened 2023-01-09
[ui] fix UI drop confirm, validate auction form, commad delimter sell…

vanminh1701:
… modal

[LIQ-1179]
[LIQ-1183]
[LIQ-1173]

[LIQ-1179]: https://liqnft.atlassian.net/browse/LIQ-1179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LIQ-1183]: https://liqnft.atlassian.net/browse/LIQ-1183?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LIQ-1173]: https://liqnft.atlassian.net/browse/LIQ-1173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/ui/src/components/SellModal/index.tsx
+++ b/core/ui/src/components/SellModal/index.tsx
@@ -28,6 +28,15 @@ export type SellModalProps = {
   sell: (nft: SingleTokenInfo, price: number) => Promise<string>;
 };
 
+const getLocalNumber = (price: string) => {
+  const arr = price.split('.');
+
+  if (price.includes('.') && arr[1] === '') {
+    return `${arr[0].replace(/\d(?=(?:\d{3})+$)/g, '$&,')}.`;
+  }
+  return arr[0].replace(/\d(?=(?:\d{3})+$)/g, '$&,') + (arr[1] ? `.${arr[1]}` : '');
+};
+
 export const SellModal: React.FC<SellModalProps> = ({
   onCancel: onUnSelectItem,
   nft,
@@ -40,7 +49,7 @@ export const SellModal: React.FC<SellModalProps> = ({
   getTokenMetadataByMintAddress,
   sell
 }) => {
-  const [formState, setFormState] = useState<{ price: number | undefined }>({
+  const [formState, setFormState] = useState<{ price: string | undefined }>({
     price: undefined
   });
   const [state, setState] = useState(TransactionState.DISPLAY);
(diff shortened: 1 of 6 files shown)


core/ui/src/components/SellModal/index.tsx
```diff
@@ -81,15 +90,20 @@ export const SellModal: React.FC<SellModalProps> = ({
 
   // Check active button submit
   const onChangeInput = (e: React.ChangeEvent<HTMLInputElement>) => {
-    if (e.target.value === '') {
+    let price = e.target.value.replace(/,/gi, '');
```
RustySol, line 93: is it save to assume `e.target.value` is a `string`?

core/ui/src/components/SellModal/index.tsx
```diff
@@ -81,15 +90,20 @@ export const SellModal: React.FC<SellModalProps> = ({
...
+    if (price === '') {
       setFormState((f) => ({ ...f, price: undefined }));
       return;
     }
 
-    const regex3Decimals = new RegExp('^[0-9]{1,11}(?:.[0-9]{1,3})?$');
-    if (regex3Decimals.test(e.target.value)) {
-      setFormState((f) => ({ ...f, price: +e.target.value }));
+    const arr = price.split('.');
+    if (arr[0].length > 11) return;
```
RustySol, line 101: can you explain what this is for?

core/ui/src/components/SellModal/index.tsx
vanminh1701: the value from `onChange` event is always `string` type even though you're using input number. and because the requirement is show commas for thound number, so need to save it as string and add preprocess function before rendering and save to local state

core/ui/src/components/SellModal/index.tsx
vanminh1701: it is because we have requirement with the number of fraction is 11, decimal is 3. you can refer the `regex` pattern on the old code

RustySol (approved):

PR #525, opened 2023-01-13
[ui][sdk] fix overflow price value

vanminh1701:


core/sdk/src/shop/eth/EthCandyShop.ts
```diff
@@ -132,7 +132,11 @@ export class EthCandyShop extends BaseShop {
     const shop = shopResult.result;
 
     const offerer = await wallet.getAddress();
-    const priceValue = new Decimal(`${price}e${this.currencyDecimals}`).toString();
+    const priceValue = ethers.BigNumber.from(price * 1000)
```
RustySol, line 135: the reason why we `*1000` and then `/1000` is to avoid the numbers that might potentially get rounded off?

RustySol: `ethers.BigNumber` only accept int value

RustySol (approved):

PR #534, opened 2023-01-17
[core]: specify which chain to create shop

david30907d:
# WIP

want to have your eyes on my proposal
1. add `blockchain` param in `candyShop` SDK
2. update `findShopByIdentifier`'s where clause with additional `blockchain` param
3. check subsequent logics and add `blockchain` param if needed

### DIFF
--- a/core/sdk/src/CandyShopInfoAPI.ts
+++ b/core/sdk/src/CandyShopInfoAPI.ts
@@ -90,9 +90,10 @@ export function fetchShopsByOwnerAddress(ownerAddress: string): Promise<ListBase
 export function fetchShopsByIdentifier(
   ownerAddress: string,
   treasuryMint: string,
-  programId: string
+  programId: string,
+  blockchain: string
 ): Promise<SingleBase<CandyShopResponse>> {
-  return fetchShopByIdentifier(axiosInstance, ownerAddress, treasuryMint, programId);
+  return fetchShopByIdentifier(axiosInstance, ownerAddress, treasuryMint, programId, blockchain);
 }
 
 export function fetchOrderByShopAndMintAddress(
--- a/core/sdk/src/factory/backend/ShopAPI.ts
+++ b/core/sdk/src/factory/backend/ShopAPI.ts
@@ -42,9 +42,10 @@ export async function fetchShopByIdentifier(
   axiosInstance: AxiosInstance,
   ownerAddress: string,
   mint: string,
-  programId: string
+  programId: string,
+  blockchain: string
 ): Promise<SingleBase<CandyShop>> {
-  const url = `/shop/owner/${ownerAddress}/mint/${mint}/programId/${programId}`;
+  const url = `/shop/owner/${ownerAddress}/mint/${mint}/programId/${programId}/blockchain/${blockchain}`;
   return axiosInstance.get<SingleBase<CandyShop>>(url).then((response) => response.data);
 }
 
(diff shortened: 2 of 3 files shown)


core/sdk/src/factory/backend/ShopAPI.ts
```diff
@@ -42,9 +42,10 @@ export async function fetchShopByIdentifier(
   axiosInstance: AxiosInstance,
   ownerAddress: string,
   mint: string,
-  programId: string
+  programId: string,
+  blockchain: string
```
RustySol, line 46: the type should be `enum Blockchain`

core/sdk/src/shop/eth/EthCandyShop.ts
```diff
@@ -19,6 +19,7 @@ const Logger = 'EthCandyShop';
 export interface EthShopConstructorParams extends BaseShopConstructorParams {}
 export interface EthShopInitParams extends Omit<EthShopConstructorParams, 'settings'> {
   settings: Partial<ShopSettings>;
+  blockchain: string
```
RustySol, line 22: we dont need this field, can use `BaseShopConstructorParams.env` already

core/sdk/src/CandyShopInfoAPI.ts
```diff
@@ -90,9 +90,10 @@ export function fetchShopsByOwnerAddress(ownerAddress: string): Promise<ListBase
 export function fetchShopsByIdentifier(
   ownerAddress: string,
   treasuryMint: string,
-  programId: string
+  programId: string,
+  blockchain: string
```
RustySol, line 94: the type should be enum Blockchain

core/sdk/src/factory/backend/ShopAPI.ts
```diff
@@ -42,9 +42,10 @@ export async function fetchShopByIdentifier(
...
 ): Promise<SingleBase<CandyShop>> {
-  const url = `/shop/owner/${ownerAddress}/mint/${mint}/programId/${programId}`;
```
RustySol, line 47: lets add `blockchain` conversion logic like this https://github.com/LIQNFT/candy-shop/blob/master/core/sdk/src/factory/backend/OrderAPI.ts#L35

Suggest to extract the logic out into another function in `utils.ts`

```
// Need to perform mapping because backend does not distinguish SOL in staging or production environments as opposed to the EVM chains, which have their own names.
function mapToCompatibleBlockchain(blockchain: Blockchain) {

  if (
    blockchain === Blockchain.SolDevnet ||
    blockchain === Blockchain.SolMainnetBeta
  ) {
    return Blockchain.Sol;
  }
  return blockchain;
}
```

core/sdk/src/CandyShopInfoAPI.ts
haVincy: Yeah. Should use `Blockchain` as type from `@liqnft/candy-shop-types`

- https://github.com/LIQNFT/candy-shop/blob/master/core/types/src/query/query.ts#L94

core/sdk/src/shop/eth/EthCandyShop.ts
haVincy: For your reference

- https://github.com/LIQNFT/candy-shop/blob/master/core/sdk/src/shop/base/BaseShop.ts#L37

core/sdk/src/CandyShopInfoAPI.ts
david30907d: got it~ thx for the info

core/sdk/src/factory/backend/ShopAPI.ts
david30907d: has discussed about this one offline, will create this mapping func in utils

PR #538, opened 2023-01-17
[types] edtion redemption

vanminh1701:


core/types/src/query/query.ts
```diff
@@ -13,7 +13,7 @@ export interface OrderSortBy extends SortBy {}
 
 type attributeType = { [key: string]: string };
 
-interface CommonQuery {
+export interface CommonQuery {
```
RustySol, line 16: do we need to export this?

vanminh1701: I need this for some api queries on admin repo.

PR #550, opened 2023-02-02
[feat] styling wert widget

vanminh1701:
[LIQ-1235]

[LIQ-1235]: https://liqnft.atlassian.net/browse/LIQ-1235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/ui/src/components/BuyModal/index.tsx
```diff
@@ -118,42 +132,34 @@ export const BuyModal: React.FC<BuyModalProps> = ({
...
-
-  const isProd = useMemo(() => {
-    const url = getBaseUrl(candyShopEnv);
-    return isCandyShopProdUrl(url);
-  }, [candyShopEnv]);
+  }, [getEvmOrderPayloadCallback, walletPublicKey, order]);
 
   useEffect(() => {
-    // Only fetch when haven't retrieved the creditCardPayAvailability
-    if (creditCardPayAvailable === undefined) {
```
haVincy, line 130: Are we sure we avoid calling the `CandyShopPay.checkPaymentAvailability` more than 1 time after removed this `if`?

core/ui/src/components/BuyModal/index.tsx
```diff
@@ -168,8 +174,90 @@ export const BuyModal: React.FC<BuyModalProps> = ({
...
+      ...otherWidgetOptions,
+      ...nftOptions,
+      listeners: {
+        ['position']: (data: any) => console.log(`${Logger}: step:`, data.step),
+        ['payment-status']: (data: any) => handlePaymentStatusCallback(data, uuid)
+      }
+    });
+    wertWidget.open();
+    onProcessingPay(BuyModalState.PROCESSING);
+  };
```
haVincy, line 232: We should separate the 3rd party logic in another component, keep the general BuyModal logic in BuyModal/index.tsx. So we are able to decouple it easily or extend with others in less comple

RustySol:
test link: https://feat-style-wert.d22tqykss69onq.amplifyapp.com/

core/ui/src/components/BuyModal/index.tsx
vanminh1701: I will check it, I remove cuz this is a bad pattern: Don't put state and setState in the same useEffect logic. We can avoid it by checking which dependency is making `useEffect` re-run

core/ui/src/components/BuyModal/index.tsx
vanminh1701: now the wert payment is just a function, we don't need UI for it so I think we can leave it here, if it is really need extend, can put in into a custom hook

vanminh1701: btw, why we need to get `getEvmOrderPayloadCallback` from component props ? I think we can put it inside `BuyModal.tsx` component, not sure if it is relate to your structure

core/ui/src/components/BuyModal/index.tsx
haVincy: I see, which means we should rely on dependency array as possible as we can, if can not simply rely on it, we will need such handling. 

What's the result after the check?

core/ui/src/components/BuyModal/index.tsx
haVincy: Offline synced.

core/ui/src/components/BuyModal/index.tsx
```diff
@@ -102,58 +97,63 @@ export const BuyModal: React.FC<BuyModalProps> = ({
...
+  const getEvmOrderPayload = useCallback(
+    (walletPublicKey: string) => {
+      if (getEvmOrderPayloadCallback) {
+        getEvmOrderPayloadCallback(walletPublicKey, order)
+          .then((evmPayload: OrderPayloadResponse | undefined) => {
+            console.log('debugger: orderPayload=', evmPayload);
+            setEvmOrderPayload(evmPayload);
+          })
+          .catch((error: Error) => {
+            console.log(`${Logger}: getEvmOrderPayloadCallback failed, error=`, error);
```
haVincy, line 111: We need to handle the error transition here as well when error is thrown from `getEvmOrderPayloadCallback`, otherwise, the BuyModal will show "Transaction Confirmed".

core/ui/src/components/BuyModal/useWertIo.ts
```diff
@@ -0,0 +1,138 @@
...
+          setPaymentInfo(res.result);
+        })
+        .catch((err: Error) => {
+          console.error(`${Logger}: confirmPayment on CandyShop server failed, error=`, err);
+        });
+    } else if (data.status === 'success') {
+      // Transit to Confirm UI after certain timeouts
+      setTimeout(() => {
+        onProcessingPay(BuyModalState.CONFIRMED);
+      }, 5000);
```
haVincy, line 71: We can remove the `setTimeout` here if use `.open` solution to render Wert Widget in another tab.

core/ui/src/components/BuyModal/useWertIo.ts
```diff
@@ -0,0 +1,138 @@
...
+    };
+    const wertWidget = new WertWidget({
+      ...wertConfirmInfo.signedData,
+      ...otherWidgetOptions,
+      ...nftOptions,
+      listeners: {
+        ['position']: (data: any) => console.log(`${Logger}: step:`, data.step),
+        ['payment-status']: (data: any) => onWertPaymentStatus(data, uuid)
+      }
+    });
```
haVincy, line 103: As offline discussion, we should handle other cases from available listeners from Wert Widget as possible as we can.

core/ui/src/components/BuyModal/index.tsx
vanminh1701: checked on new code, 1 run per function

core/ui/src/components/BuyModal/index.tsx
vanminh1701: evmOrderPayload isi required data when run function wert, I added logic check if this api return error, user will see Buy button is disabled

core/ui/src/components/BuyModal/useWertIo.ts
vanminh1701: tested with `close`, I don't see its trigger, leave a console for debugger,
add log for `loaded` event, but we don't need to use it

core/ui/src/components/BuyModal/index.tsx
haVincy: Ack

core/ui/src/components/BuyModal/useWertIo.ts
haVincy: Is it missed to commit? Not seeing the `loaded` event type in this `listeners` calback.

haVincy (approved):

PR #554, opened 2023-02-21
[ui] remove filter shop and add collection filter dropdown style

vanminh1701:


vanminh1701:
https://github.com/LIQNFT/candy-shop/pull/554/files#diff-17a03c4c1c4603e6ddd65739f02595760739af00965def0f58f64f1c7e4f0ee5R37 - added a optional prop to display whether dropdown style or plain list style for collection filter. Default view will be plain list

RustySol (approved):

